PR #14, opened 2013-08-21
[pom.xml] changed parent groupId

chaensel:
changed parent groupId from
com.github.johanneslichtenberger.sirix
to
com.github.sirixdb.sirix

There are more referenes to com.github.johanneslichtenberger.sirix in other pom.xml files.


JohannesLichtenberger:
@all-contributors please add @chaensel for doc

PR #26, opened 2014-05-06
Fix typo in javadoc

batista:
The javadoc was referencing `pToKey` when it should refer to `fromKey` instead


JohannesLichtenberger:
@all-contributors please add @batista for doc

PR #99, opened 2019-09-28
Actually copy `cert.pem` and `key.pem` to `/opt/sirix/sirix-data` in …

anmonteiro:
…Dockerfile

Apparently Docker needs the trailing slash to know that it needs to create the directory.

Before this commit `sirix-data` was just a file with the contents of `key.pem`

### DIFF
--- a/bundles/sirix-rest-api/Dockerfile
+++ b/bundles/sirix-rest-api/Dockerfile
@@ -20,8 +20,8 @@ EXPOSE 9443
 
 # Copy your fat jar to the container
 COPY target/$VERTICLE_FILE $VERTICLE_HOME/
-COPY src/main/resources/cert.pem $VERTICLE_HOME/sirix-data
-COPY src/main/resources/key.pem $VERTICLE_HOME/sirix-data
+COPY src/main/resources/cert.pem $VERTICLE_HOME/sirix-data/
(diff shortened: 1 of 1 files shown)


JohannesLichtenberger:
Thank you so much :-) hope you'll be able to get the HTTP-server up and running. Guess it would be much easier with a working docker-compose file. I'm currently running the stuff without Docker as you might have guessed ;-)

anmonteiro:
Very early stages playing with SirixDB, but I've got the keycloak server and the docker container running right now. Haven't issued any requests yet though!

JohannesLichtenberger:
Very cool. Maybe, once you're going to update a resource in a database you'll first have to get the anchor node or do a HEAD request to retrieve a hashCode of the node via the ETag HTTP-header response and send this again with the updating request (also in the ETag HTTP-Header). That way we can detect concurrent and parallel changes between the requests.

JohannesLichtenberger:
There's a paper about this from Sebastian:

http://kops.uni-konstanz.de/bitstream/handle/123456789/6102/paper.pdf?sequence=1&isAllowed=y

JohannesLichtenberger:
BTW: You can try my integration tests (written with the Vert.x HTTP-Client in Kotlin).

JohannesLichtenberger:
@all-contributors please add @anmonteiro for projectManagement

Issue #90, opened 2019-09-27
Create a favicon


JohannesLichtenberger:

themayurkumbhar:
Hi @JohannesLichtenberger 
What kind of favicon you are looking for?

JohannesLichtenberger:
Hey, good question, but I don't really like the SirixDB logo, so I guess it's totally up to you... But for https://sirix.io :-)

JohannesLichtenberger:
Hey @karmakaze, would it be possible to get a red version? :-)

karmakaze:
Absolutely. might also try out styling a bit more too.

JohannesLichtenberger:
Cool :-) thanks so much

PR #103, opened 2019-10-02
Add favicon.ico and test index.html

karmakaze:
I know the issue #90 was to create just a favicon.ico but in its comments it also mentioned not liking the existing logo so much.

I thought about it for a while: what can represent a time-travelling time-series database?

An idea that came to mind were time-travel (duh?) as in travelling faster than the speed of light which is limited by the light cone. Some [nice pictures of light cones](http://www.preposterousuniverse.com/blog/2011/09/24/can-neutrinos-kill-their-own-grandfathers/comment-page-2/) and travelling in just time or just space.

Other ideas were clocks, hourglass sand timers, space in general.

I combined a stylized 'S' in to look like the conjoined V's of a light cone and capped the ends to look like an hourglass. The openings in the edges of the light cone represents how the normal inside-the-light-cone is connected to the outer part which is normally inaccessible. A slightly purple-ish blue because space, not a complete vacuum, but energetic.

I created 16x16 and 32x32 layers in `favicon.ico` with an `index.html` to show it, and saved a screenshot of a Chrome tab in `example.png`. Link to [example.png](https://github.com/karmakaze/sirix/blob/master/example.png) on my github. Normally this would be half the width and height as it was captured on a HiDPI display.


JohannesLichtenberger:
Wow, thanks so much. I really like it :-)

PR #105, opened 2019-10-02
In red with a bit of rounding on the top/bottom 'cone ends'

karmakaze:
The idea was to make it look more 'conical' and rounding out the ends gives it more of a figure-8 `infinity` feel to it!

What do you think of the shade of red?

I pulled it out from a section of the previous logo. I also like the fuchsia-red on the outer ring on [this page](https://dev.to/johanneslichtenberger/sirixdb-importance-of-efficiently-retaining-the-history-of-your-data-5c1m).

example


JohannesLichtenberger:
@all-contributors please add @karmakaze for design

PR #106, opened 2019-10-02
Java9 modules

b0n541:
Hi,

this is a preliminary pull request. I managed to introduce Jigsaw modules and in my IDE (read IntelliJ) everything is working. 

But Maven is not able to compile the test sources of module sirix.xquery because it can't resolve the base test classes from module sirix.core.

I followed the approach from this example repository: https://github.com/jhenriquez418/java-cross-module-testing/blob/master/pom.xml

I don't see any differences to the example project. Maybe you have an idea?

Many thanks,

Jan


JohannesLichtenberger:
Hmm, I'm not sure. It might be because the test case base classes are directly in org.sirix which is not exported in the core module?

JohannesLichtenberger:
Thank you so much for all the work :-)

b0n541:
I already tried to move the test base classes into the org.sirix.util package with no success. In the example project https://github.com/jhenriquez418/java-cross-module-testing/blob/master/pom.xml the reused classes are also in a package that is not exported via Jigsaw.

JohannesLichtenberger:
Maybe because it's in `src/main/java` and the tests are in `src/test/java`? Can you try if it works if you add a module there and export the utils package in your fork? Hmm, I have to read more about the modules system I think. Would be cool to get this working before I merge your stuff :-) maybe I could also ask for help on `dev.to` if it's still not working. What do you think?

JohannesLichtenberger:
Forget my suggestion regarding a different module. It's probably a Maven (surefire) plugin issue:

https://stackoverflow.com/questions/46613214/java-9-maven-junit-does-test-code-need-module-info-java-of-its-own-and-wher

b0n541:
I have seen that Stackoverflow discussion as well. At the moment the test-compile step in bundle sirix-xquery is failing because it can't resolve classes from the core bundle. So I think it's not a Surefire plugin issue it's rather a compiler plugin thing. My knowledge about Maven is a bit rusty  lately because I use Gradle most of the time. So it might be a good idea to ask on dev.to.

JohannesLichtenberger:
Hm, as of now no ideas it seems!?

JohannesLichtenberger:
Here: https://dev.to/johanneslichtenberger/help-with-maven-multi-module-project-java-module-system-with-a-pr-3ifi

b0n541:
I saw that you accepted the migration to Gradle. Shall I try again with Gradle to add the Java Jigsaw modules?

JohannesLichtenberger:
Hm, the Gradle stuff is not finished, but when it is, yes :-)

JohannesLichtenberger:
Hey, I just merged another PM a couple of days ago. Still need to work on some Gradle stuff, but if you still like you could try again :-)

JohannesLichtenberger:
@b0n541 you could definitely try again, if you have some time.

Issue #100, opened 2019-09-28
Starting a rest-api service that fails to connect to keycloak exits with status code 0


anmonteiro:
<details>
<summary>Full Stack Trace</summary>

```
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:8080
        at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
        at java.base/sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:779)
        at io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:327)
        at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:340)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:632)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:579)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:496)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:458)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:897)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.net.ConnectException: Connection refused
        ... 11 more
Sep 28, 2019 11:28:52 PM io.vertx.core.impl.launcher.commands.VertxIsolatedDeployer
SEVERE: Failed in deploying verticle
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:8080
        at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
        at java.base/sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:779)
        at io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:327)
        at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:340)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:632)
[...]

Issue #51, opened 2018-12-27
Docker compose


JohannesLichtenberger:
As the RESTful-API and thus the docker container is needing a started Keycloak instance, we could provide a Docker Compose file.

kmjayadeep:
Can I take this up?

JohannesLichtenberger:
Sure, thank you :-)

Maybe you can talk to @anmonteiro, he found an issue (I think he might have added Keycloak already, but I'm not sure :)):
https://github.com/sirixdb/sirix/issues/100

Issue #50, opened 2018-12-27
Docker setup / Fix README


JohannesLichtenberger:
- pull the image from Docker Hub
- create a docker container
- update the container with an adapted sirix-conf.json file (client.secret value for Keycloak auth must be set and maybe the listening port for the HTTP(S) verticle
- start the docker container

Make it work, I'm by no means sadly no docker guru.. and fix the README section accordingly.

PR #108, opened 2019-10-03
[WIP] Docker and docker-compose issues

kmjayadeep:
Working on 
#50 Docker setup / Fix README
#51  Docker compose
#100 Starting a rest-api service that fails to connect to keycloak exits with status code 0

Some initial feedback would be great before proceeding further, as I have limited knowledge about the project.

1. Planning to add multi stage docker build, so that we can avoid doing mvn build before running docker
2. Added image name to the docker compose file itself, Avoid running docker commands directly.
3. Network configuration is unnecessary with docker compose. Containers are reachable to each other with the service name as hostname. Initially we could provide in the readme to change the hostname in sirix-conf.json file as:
```"keycloak.url": "http://keycloak:8080/auth/realms/master"```
Eventually I would suggest using environment variables and .env file instead of conf file itself.

The volume mount seems to be a bit confusing. (Sorry again for the lack of in depth knowledge about the project). Please review and suggest if any changes are required in volume mounts I have changed. Absolute paths in host system is not recommended as it won't be platform specific.

### DIFF
--- a/bundles/sirix-rest-api/docker-compose.yml
+++ b/bundles/sirix-rest-api/docker-compose.yml
@@ -2,6 +2,7 @@ version: '3'
 services:
   web:
     build: .
+    image: sirixdb/sirix
     ports:
       - "9443:9443"
     volumes:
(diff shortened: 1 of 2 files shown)


JohannesLichtenberger:
Thank you so much :-)

To make a simple test configuration, maybe we can add the Keycloak export realm data:

https://github.com/sirixdb/sirix/blob/master/bundles/sirix-rest-api/src/test/resources/realm-export.json

I think it should be `KEYCLOAK_IMPORT`.

PR #113, opened 2019-10-05
107 migrate to gradle

LarsEckart:
it's still work in progress, but that way you can have a look and point out things that don't work yet.

So far I only ran gradle init to get the gradle wrapper and build.gradle files into each module. sirix-core also uses the shadowjar plugin to build a fat jar. Especially for this I think I'll need your feedback, if the artefacts are all built correctly.


LarsEckart:
reminder to myself: once it's working, removing all traces of maven from the project also still needs to be done :)

LarsEckart:
is aspectj actually used in the project? I found the maven plugin in a pom and also the dependency, but no code has any imports from org.aspectj :/

JohannesLichtenberger:
Ah no, AspectJ is not needed IIRC.

LarsEckart:
Johannes, I wasn't done yet :D the pipeline for example still uses maven and also all the maven files are still around and we should test that the created artefacts match what maven was building ;)

JohannesLichtenberger:
Yes, but I think it doesn't matter if the Gradle files are already there ;-)

JohannesLichtenberger:
Hey, had to remove the Gradle stuff temporarily... are you still working on it? :+1:

JohannesLichtenberger:
@all-contributors please add @LarsEckart for code

PR #120, opened 2019-10-15
Make OAuth flow configurable between PASSWORD and AUTH_CODE

MrBuggySan:
- Make OAuth flow configurable between PASSWORD and AUTH_CODE
- add new endpoint for getting the auth server url login page with redirect_uri set
- Change name of /login endpoint to /token 
- add relevant values in config file

I tested the auth_flow manually with my own web app. If there is a way to make an automated test based on this, let me know. I'm curious how to do that.  

Please send me your feedback on what you think

### DIFF
--- a/bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
+++ b/bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
@@ -2,6 +2,7 @@ package org.sirix.rest
 
 import io.netty.handler.codec.http.HttpResponseStatus
 import io.vertx.core.http.HttpServerResponse
+import io.vertx.core.json.JsonObject
 import io.vertx.core.net.PemKeyCertOptions
 import io.vertx.ext.auth.oauth2.OAuth2FlowType
 import io.vertx.ext.web.Route
@@ -17,6 +18,7 @@ import io.vertx.kotlin.ext.auth.authenticateAwait
 import io.vertx.kotlin.ext.auth.oauth2.oAuth2ClientOptionsOf
 import io.vertx.kotlin.ext.auth.oauth2.providers.KeycloakAuth
 import kotlinx.coroutines.launch
+import org.apache.http.HttpStatus
 import org.sirix.rest.crud.CreateMultipleResources
 import org.sirix.rest.crud.Delete
 import org.sirix.rest.crud.json.*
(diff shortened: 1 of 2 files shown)


JohannesLichtenberger:
Hey, that's super helpful. Thank you very much. Will be used for the web frontend for sure :-)

JohannesLichtenberger:
I have to test this somehow automatically... but does your web application still work? I changed a few things yesterday... so, what I currently don't know is if Vert.x does this exchanging of the authentication code with a token automatically. But how does the browser get the token for adding it to the Authentication HTTP-Header?

JohannesLichtenberger:
I had implemented this with a SessionHandler and a CookieHandler, but would like to leave it stateless :-)

MrBuggySan:
The front end has to know if there is a token in its cache somewhere. If not it has to ask sirix-rest-api for the login page of keycloak via GET /user/authorize. 

sirix-rest-api returns a redirect 302 with the url of the keycloak login page (embedded on it is  the redirect url to the front end page route to handle the rest of the process). User enters their credentials then is redirected to redirect url that was embedded before.

The front end page route will then handle grabbing the auth_code, which it will then exchange for a token via the GET /token endpoint of sirix-rest-api. 

I figure with with this AUTH_CODE flow, there must always be logic in frontend to handle extracting the auth_code and exchanging it with the token. I'm not sure how all this can be stateless? Let me know what your thoughts.

MrBuggySan:
This is the route the user would redirect to after logging in:
https://gist.github.com/MrBuggySan/8d8b26a80ed0d34cf1b457d9cde73bf1

JohannesLichtenberger:
Ah yes, I think if we provide a "/callback" route (instead of redirecting to the actual page), then Vert.x is getting the code itself from the URL-parameter and exchanging it with the token. Then it's injecting this token into a User and we could for instance use a SessionHandler and a CookieHandler, so the token should be valid for the whole session I guess (but it's a bit Vert.x magic).

But I think our approach right now is much more RESTful (extracting the code on the frontend, then POSTing it to the token endpoint to exchange the auth code with the token :-) I have to read why in this OAuth2 authentication code flow for security reasons first an auth code is sent, which then has to be exchanged again.

And thanks for your Gist... will add this to our [Web Frontend](https://github.com/sirixdb/sirix-web-frontend) then (using Vue.js/Nuxt.js) :-)

JohannesLichtenberger:
@all-contributors please add @MrBuggySan for code

Issue #87, opened 2019-09-26
Playing with the APIs


JohannesLichtenberger:
Checking what makes and makes no sense... finalize APIs.

haridevelops:
what is this issue all about ? can you please eloborate ?

JohannesLichtenberger:
If we need to change minor things. Method names, signatures...

FernandaCG:
Hi! Can I work on this?

JohannesLichtenberger:
Sure :-)

gourcool:
@JohannesLichtenberger Can i work on this

PR #150, opened 2019-12-30
Refactor in rest-api

FernandaCG:
Change in signatures and variables according to #87


JohannesLichtenberger:
That's helpful. Can you however update your branch and maske the PR again? Seems we have some conflicts.

Thanks a lot, Fernanda :-)

PR #151, opened 2019-12-31
Refactor in rest-api

FernandaCG:
Change in signatures and variables according to #87


JohannesLichtenberger:
@all-contributors please add @FernandaCG for code

Issue #128, opened 2019-10-28
Use TestContainers for integration tests


yiss:
The problem now is that when running integration tests we need to have a instance if keycloak running (the documentation shows how to run it using Dokcer)

Ideally we want to run integration tests without having to rely on having an instance of Keyloak running. To achieve that we would like to use TestContainers (official documentation : https://www.testcontainers.org/) to run Keycloak

Some other resources that may help : 
  - https://www.7p-group.com/blog/integrationstests-mit-testcontainers-spring-security-und-keycloak/
  - https://github.com/cremich/testcontainers-keycloak

@JohannesLichtenberger Can you export realm-export.json ?

JohannesLichtenberger:
A realm export can be found here:

https://github.com/sirixdb/sirix/blob/master/bundles/sirix-rest-api/src/test/resources/realm-export.json

Thanks for adding this issue :-)

JohannesLichtenberger:
Another resource (in german however):

https://www.7p-group.com/blog/integrationstests-mit-testcontainers-spring-security-und-keycloak/

kulovan:
Hi, can I jump on this issue?

balajiv113:
@JohannesLichtenberger 
Is this still up ? I could see a reverted commit in log.

JohannesLichtenberger:
Hmm, maybe we can run the docker container via travis and add a user with the appropriate roles afterwards.

yiss:
I've started working on it, while ago, hadn't had the chance to finish it. If you want I can give you a lead on what I've already did

JohannesLichtenberger:
@yiss I reverted my changes regarding TestContainers, because I think starting a Keycloak container plus adding a user with the rights can be simply done via Travis:

https://docs.travis-ci.com/user/docker/

balajiv113:
@JohannesLichtenberger we could also run docker image via gradle before test and after test we can stop the container

https://github.com/bmuschko/gradle-docker-plugin

yiss:
@JohannesLichtenberger yeah that is more elegant implementation, maybe add the @balajiv113 suggestion too so that other contributers don't have to configure manually docker with keykloack

PR #180, opened 2020-01-09
Support for automated rest-api integration test - #128

balajiv113:
Automated Rest API Integration test using docker and gradle.

Before test docker-compose will be started and after test docker-compose will be shutdown

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ jobs:
       name: "Source Build"
       jdk: openjdk13
       install: skip
-      script: ./gradlew clean build --info
+      script: ./gradlew clean build --info --stacktrace
       after_success:
         - ./gradlew jacocoRootReport coveralls
     - stage: deploy
--- a/build.gradle
+++ b/build.gradle
@@ -186,8 +186,7 @@ subprojects {
     test {
         useTestNG()
         useJUnit()
-        //To run Kotlin Juint once it is fixed
-        //useJUnitPlatform()
+        useJUnitPlatform()
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -186,7 +186,6 @@ subprojects {
     test {
         useTestNG()
         useJUnit()
-        useJUnitPlatform()
     }
 }
 
--- a/bundles/sirix-rest-api/build.gradle
+++ b/bundles/sirix-rest-api/build.gradle
@@ -34,6 +34,7 @@ dependencies {
     testImplementation testLibraries.kotlinTestJunit
     testImplementation testLibraries.junit
     testImplementation testLibraries.vertxJunit5
+    testImplementation testLibraries.jsonassert
 }
 
 description = 'Non-blocking REST-API for SirixDB.'
--- a/bundles/sirix-rest-api/build.gradle
+++ b/bundles/sirix-rest-api/build.gradle
@@ -1,16 +1,19 @@
 buildscript {
     repositories {
         mavenCentral()
+        jcenter()
     }
 
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.61"
+        classpath "com.avast.gradle:gradle-docker-compose-plugin:0.10.7"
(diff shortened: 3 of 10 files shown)


JohannesLichtenberger:
Hm, we have to investigate what's going wrong especially with these two:

org.sirix.rest.SirixVerticleXmlTest > testGet(Vertx, VertxTestContext) FAILED

    java.util.concurrent.TimeoutException: The test execution timed out. Make sure your asynchronous code includes calls to either VertxTestContext#completeNow(), VertxTestContext#failNow() or Checkpoint#flag()

org.sirix.rest.SirixVerticleXmlTest > testPut(Vertx, VertxTestContext) FAILED

    java.util.concurrent.TimeoutException: The test execution timed out. Make sure your asynchronous code includes calls to either VertxTestContext#completeNow(), VertxTestContext#failNow() or Checkpoint#flag()

I think the other one needs a simple lexical sorting of database names first. Succeeded locally, but I guess it depends on the filesystem...

balajiv113:
Ya agreed. Even the 1st 2 were successful in my local machine. All tests were successful locally.

JohannesLichtenberger:
Yeah, for me too the last time I executed them... but I can only think that the directory listing might be different depending on the filesystem. So, sorting by directory name first will probably fix this one.

The other two, hmmm.

balajiv113:
@JohannesLichtenberger 
I have fixed the JSON difference and also the timeout issue. The one new issue that is present is testDeleteResource which is giving 500 Internal Server error. That to in random not in all builds.

JohannesLichtenberger:
If you disable the test, the other tests run every time without an assertion error? Hm, usually the exception should get printed, but I couldn't see anything. Maybe we can disable the test for now if it otherwise runs every time fine. I'd love to merge :-)

JohannesLichtenberger:
So, the other tests were failing because Keycloak wasn't up, right?

balajiv113:
Yes thats correct, the timeout exception were due to keycloak not fully up. To solve that i added custom health check in docker-compose.

JohannesLichtenberger:
Ah cool. I just checked the code and I couldn't see an obvious bug when a resource in a database should be deleted, hmm. That said, I think sometimes the exceptions don't pop up...

balajiv113:
As you could see, in travis this time that testDeleteResource is successful. May be will check for one more build if it fails again will disable it for now and analyse it further.

JohannesLichtenberger:
Hm, probably we can also increase the timeout in the health check. I think the Keycloak start is rather slow.

JohannesLichtenberger:
Damn, still fails

balajiv113:
Ahh ya. In my local system its not failing if that fails then can debug easily and fix it.

JohannesLichtenberger:
Same here, not failing locally. But it would be easy to fix if we get the stacktrace, hmm.

JohannesLichtenberger:
I think my Vert.x error handling is not always right, somehow...

JohannesLichtenberger:
I think at first delete(...) has to be called before each test and after each test to cleanup stuff. I'm not sure what's now failing, though.

balajiv113:
@JohannesLichtenberger Looks like in travis we need to disable docker deploy for pull requests

balajiv113:
> I think at first delete(...) has to be called before each test and after each test to cleanup stuff. I'm not sure what's now failing, though.

No i tried with that, delete actually deletes the HTTP server started by vertx for testing so we will be getting connection refused for all the test cases

JohannesLichtenberger:
> @JohannesLichtenberger Looks like in travis we need to disable docker deploy for pull requests

Yes, absolutely

JohannesLichtenberger:
> > I think at first delete(...) has to be called before each test and after each test to cleanup stuff. I'm not sure what's now failing, though.
> 
> No i tried with that, delete actually deletes the HTTP server started by vertx for testing so we will be getting connection refused for all the test cases

Yes, saw this happening. I disabled the test for now and I'm going to merge. I hope we'll figure out why the test fails on Travis.

PR #186, opened 2020-01-14
Docker compose updates with create-keycloak startup script

balajiv113:
I have updated the docker-compose to use keycloak startup-scripts. This way we can create any type of user with any type of user before start-up of keycloak server.

**Problems**
1. Docker compose V3 doesn't support depends_on condition. So sirix_server will fail until the keycloak is up and running.
**Soultion** 
Since `restart: always` is set, sirix_server will restart always until keycloak is up
2. sirix-config.json contains keycloak URL as localhost:8080 / localhost:3005. Keycloak ports is not bound to sirix container thus 8080 and 3005 will not be accessible. 
**Soultion**
Modified localhost to keycloak (service name). Such that keycload will be resolved to keycloak server IP.


JohannesLichtenberger:
Very nice, thanks :-)

PR #200, opened 2020-02-13
Refactor compare prefixes

add09:
Refactor the Calc class to remove unnecessary code repetition by replacing it with one function or calling an overloaded one.

Also wrote tests for Calc.compare prefix functions.


JohannesLichtenberger:
Can you make a PR with only your changes? :-)

PR #201, opened 2020-02-13
Bugfix/JsonNumber. stringToNumber

mikkar2:
The function was unable to return a double. Reason is:
number = Float.valueOf(stringValue);

Float.valueOf() will always round String to a float. So when a value is to big for a float it will get rounded to a float and return.

The call: number = Double.valueOf(stringValue); will never be used

Bug is fixed and 6 testcases are added.


JohannesLichtenberger:
Wow, great, thanks a lot :)

JohannesLichtenberger:
@all-contributors please add @mikkar2 for code

PR #203, opened 2020-02-13
refactor: Removed code repetion in the calc class and created tests f…

add09:
…or compare prefix, int and long.

Removed unnecessary import


JohannesLichtenberger:
@all-contributors please add @add09 for code

PR #220, opened 2020-02-27
tests: extend StorageTest and fix bug in RAMStorage.truncate

EmilGedda:
The underlying bool in `RAMStorage.exists()` is only set on construction, and not properly set to false when `.truncate()` is called. This patch sets the bool to false in `truncate` and then sets it to true whenever `write` and `writeUberPageReference` succeeds. The patch also adds tests for the bug.

The whitespace noise comes from `RAMStorage.java` having Windows CRLF line endings. I took the liberty to change this to Unix line endings to match the rest of the source tree.


JohannesLichtenberger:
Thanks a lot :+1:  I have to admit, that I haven't even used the RAM-only storage ;-)

In the next months I'll concentrate on cleaning up code and writing unit and integration tests. I guess, that a compile time DI-framework like Dagger2 would be very useful.

PR #237, opened 2020-04-20
Coroutine axis

marcinbieleckiLLL:
New bundle: sirix-kotlin-api with CoroutineAxis - ConcurrentAxis rewriten to kotlin coroutines


JohannesLichtenberger:
Thanks a lot. BTW: If you like to you can get rid of these "m"-prefixes for member/instance variables. I want to get rid of them in the whole project.

Oh, and if you like to you can write a benchmark. Would also be interesting how this performs on modern hardware for larger datasets.

JohannesLichtenberger:
@all-contributors please add @marcinbieleckiLLL for code

PR #264, opened 2020-05-30
Closes #191

### LINKED ISSUE #191: Making keeping track of the childCount of each node configurable
To further reduce storage space we can make it optional to keep track of the `childCount` of nodes.

punit-kulal:
closes #191 
Following changes were made
- Introduced a parameter in ResouceConfig
- Start respecting the parameter in JsonNodeTrxImpl and XmlNodeTrxImple
- Changes to respect parameter while serializing and deserializing StructNodeDelegate


JohannesLichtenberger:
Great work, if you like to you could do the same for the descendantCount, as it's currently tightly coupled with building hashes (as in both cases we have to adapt ancestor nodes). However, It's a little bit more tricky as if both params are set ancestor nodes should only be accessed once for setting both the hashes and the descendantCount.

PR #292, opened 2020-07-26
252 create a kotlin cli

ManfredNentwig:


JohannesLichtenberger:
Thanks a lot, that's a bunch if work you've done :)

JohannesLichtenberger:
@all-contributors please add @ManfredNentwig for code

JohannesLichtenberger:
Maybe we can extract some classes you're using from the rest-api module into a sirix-kotlin-common module/bundle :-)

ManfredNentwig:
Thank you for merging.

I would really prever to have a  common module or to implement something like a high level api wich are used by REST and CLI.
God luck with your debug session!

Issue #294, opened 2020-07-27
Insert lastChildPointer for JsonNodeTrx at least


JohannesLichtenberger:
Thus, we can provide efficient insertions into objects (always at the last position) as well as avoid triggering of relabeling DeweyIDs.

Raj-Datta-Manohar:
Hi @JohannesLichtenberger , I am new here and trying to work on this issue.  
I am stuck here at https://github.com/sirixdb/sirix/blob/81bdaaaf7931cdf9651e9b502c16df63b0274b53/bundles/sirix-core/src/main/java/org/sirix/access/trx/node/json/JsonNodeTrxImpl.java#L1616

trying to figure out if there will be a condition when inserting as Last Child similar to
https://github.com/sirixdb/sirix/blob/81bdaaaf7931cdf9651e9b502c16df63b0274b53/bundles/sirix-core/src/main/java/org/sirix/access/trx/node/json/JsonNodeTrxImpl.java#L1627-L1629

Would be great if you could help!!

JohannesLichtenberger:
Whenever a node has no right sibling I'd say, the parent must be adapted. Furthermore if a node during deletes has no right siblings the parent must be adapted.

JohannesLichtenberger:
For the current tree-encoding: https://www.baeldung.com/sirix#tree-encoding-in-sirix

Raj-Datta-Manohar:
@JohannesLichtenberger, Could you help me out in the implementation of `insertValue(...)` method for inserting an ObjectRecord as Last Child  
As of now, the `insertValue(...)` method, https://github.com/sirixdb/sirix/blob/81bdaaaf7931cdf9651e9b502c16df63b0274b53/bundles/sirix-core/src/main/java/org/sirix/access/trx/node/json/JsonNodeTrxImpl.java#L581
always inserts as the first child,  
  
So I guess I need to implement a new method `insertValueForLastChild(...)`?

JohannesLichtenberger:
@Raj-Datta-Manohar, that's a method internally used for inserting a value for a key or field (synonyms). That is, basically you only have a first child pointer and for the value a parent pointer (even though the parent pointer could be even omitted, strictly speaking just for convencience).

That is, the first child is also the last child, regarding the pointers.

Raj-Datta-Manohar:
Thanks for the clarification @JohannesLichtenberger , I will raise a PR

JohannesLichtenberger:
Thanks a lot, that's fast :-)

[...]

Issue #255, opened 2020-05-21
JsonNodeTrx / insert nodes as left sibling


JohannesLichtenberger:
Inserting nodes as a left sibling should be possible for convenience.

rbhatmanjunath:
Hi, can I work on this one?

JohannesLichtenberger:
Sure :-)

JohannesLichtenberger:
@rbhatmanjunath how is it going? Do you need help? :)

JohannesLichtenberger:
@rbhatmanjunath, let me know if you still like to work on this, please.

Issue #299, opened 2020-08-01
Transaction group commit


JohannesLichtenberger:
To avoid write amplification for small sized read-write transactions:

- we should preCommit the data during a commit to a queue
- the data should be serialized, however to the queue instead of a file
- the queue should be synced to durable storage after a specific number of bytes has been accumulated in the queue
- the RevisionRootNode plus update up to the UberPage ancestor have to be written by the code, which implements the group commit
- only then, the transaction is fully committed and the transactions in the transaction get notified and their commit methods can return a success to the callers
- the changes should still be tracked in a JSON file: the author and commit message should be tracked there. As long as a transaction group has not been synced new data is appended to the same file in the same revision.

=> a snapshot represents a group of transactions and the change tracking files track the optional author, optional message and the root nodes of the modified data plus context node infos...

Open question: how to implement time based and max number of nodes based auto-commits, as the data is not committed. Possible solution: It must be tracked in the resource manager instead of auto-committing in the transaction itself and the resource manager has to trigger the auto-commits.

JohannesLichtenberger:
Alternative and maybe better and easier approach is to write XML or JSON strings as well the context node keys and the insertion as right sibl/left sibli or first child into a queue and issue a transaction commit once a threshold is hit (node based, byte size based, time span based or a combination thereof). The queue must be replaced with a new empty queue atomically.

API-wise the open issue would be how to group the updates, such that it can be attributed to a single person (optional user and commit timestamp).

Issue #296, opened 2020-07-28
Performance: Many small writes within a transaction via insertSubtreeX(...), if committed are horribly slow


JohannesLichtenberger:
```java
    final var resource = "smallInsertions";

    try (final var database = JsonTestHelper.getDatabase(PATHS.PATH1.getFile())) {
      database.createResource(ResourceConfiguration.newBuilder(resource)
                                                   .storeDiffs(false)
                                                   .hashKind(HashType.NONE)
                                                   .buildPathSummary(false)
                                                   .build());
      try (final var manager = database.openResourceManager(resource); final var wtx = manager.beginNodeTrx()) {
        System.out.println("Start inserting");

        final long time = System.nanoTime();

        wtx.insertArrayAsFirstChild();

        var jsonObject = """
            {"item":"this is item 0", "package":"package", "kg":5}
            """.strip();

        wtx.insertSubtreeAsFirstChild(JsonShredder.createStringReader(jsonObject));

        for (int i = 0; i < 650_000; i++) {
          System.out.println(i);
          jsonObject = """
              {"item":"this is item %s", "package":"package", "kg":5}
              """.strip().formatted(i);

          wtx.insertSubtreeAsRightSibling(JsonShredder.createStringReader(jsonObject));
        }

        wtx.commit();

        System.out.println("Done inserting [" + (System.nanoTime() - time) / 1_000_000 + "ms].");
      }
    }
```

[JUnitStarter-2020-07-28.zip](https://github.com/sirixdb/sirix/files/4990616/JUnitStarter-2020-07-28.zip)

JohannesLichtenberger:
Some might be attributed to running from JUnit within IntelliJ plus Profiling, but still way too slow.

JohannesLichtenberger:
We can add a parameter to disable the `getParentKind()` check on purpose. Furthermore I wonder if it's best to share one Writer and Reader in the `ResourceManager` to make the `reinstantiate()` call during a commit as the last action more performant. Another thing might be to disable reading all nodes of the `PathSummary` into main memory during writes.

Discussions are welcome :-)

JohannesLichtenberger:
It's on SSDs and Hard drives mainly due to the fact, that we don't group transaction commits, but immediately write them to persistent storage. However, this should be okay, even for small writes on byte addressable NVM.

PR #300, opened 2020-08-05
Added lastChild and leftSibling pointer for JsonNodeTrx

Raj-Datta-Manohar:
Resolves issues #294 and #255

### DIFF
--- a/bundles/sirix-core/src/main/java/org/sirix/node/delegates/StructNodeDelegate.java
+++ b/bundles/sirix-core/src/main/java/org/sirix/node/delegates/StructNodeDelegate.java
@@ -49,6 +49,9 @@ public class StructNodeDelegate extends AbstractForwardingNode implements Struct
   /** Pointer to the first child of the current node. */
   private long firstChild;
 
+  /** Pointer to the last child of the current node. */
+  private long lastChild = -2L;
+
   /** Pointer to the right sibling of the current node. */
   private long rightSibling;
 
@@ -87,6 +90,31 @@ public class StructNodeDelegate extends AbstractForwardingNode implements Struct
     this.descendantCount = descendantCount;
   }
 
+  /**
+   * Constructor.
+   *
+   * @param nodeDelegate {@link NodeDelegate} instance
+   * @param firstChild first child key
+   * @param lastChild last child key
+   * @param rightSibling right sibling key
+   * @param leftSibling left sibling key
+   * @param childCount number of children of the node
+   * @param descendantCount number of descendants of the node
+   */
+  public StructNodeDelegate(final NodeDelegate nodeDelegate, final long firstChild, final long lastChild, final long rightSibling,
+      final long leftSibling, final @Nonnegative long childCount, final @Nonnegative long descendantCount) {
+    assert childCount >= 0 : "childCount must be >= 0!";
+    assert descendantCount >= 0 : "descendantCount must be >= 0!";
+    assert nodeDelegate != null : "del must not be null!";
+    this.nodeDelegate = nodeDelegate;
+    this.firstChild = firstChild;
+    this.lastChild = lastChild;
+    this.rightSibling = rightSibling;
+    this.leftSibling = leftSibling;
+    this.childCount = childCount;
+    this.descendantCount = descendantCount;
+  }
+
   @Override
   public NodeKind getKind() {
     return nodeDelegate.getKind();
@@ -95,6 +123,11 @@ public class StructNodeDelegate extends AbstractForwardingNode implements Struct
   @Override
   public boolean hasFirstChild() {
     return firstChild != Fixed.NULL_NODE_KEY.getStandardProperty();
+  }
+
+  @Override
+  public boolean hasLastChild() {
+    return lastChild != Fixed.NULL_NODE_KEY.getStandardProperty();
   }
 
   @Override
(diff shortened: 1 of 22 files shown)


Raj-Datta-Manohar:
@JohannesLichtenberger , Take a look at this PR:-)
I will raise a seperate PR for tests, a little busy this week.

bundles/sirix-core/src/main/java/org/sirix/node/delegates/StructNodeDelegate.java
```diff
@@ -154,15 +197,22 @@ public void incrementChildCount() {
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(childCount, nodeDelegate, firstChild, leftSibling, rightSibling, descendantCount);
+    return ((lastChild==-2L) ? Objects.hashCode(
```
JohannesLichtenberger, line 200: What's this -2L key?

bundles/sirix-core/src/main/java/org/sirix/node/delegates/StructNodeDelegate.java
```diff
@@ -154,15 +197,22 @@ public void incrementChildCount() {
...
+      childCount, nodeDelegate, firstChild, lastChild, leftSibling, rightSibling, descendantCount));
   }
 
   @Override
   public BigInteger computeHash() {
     final Funnel<StructNode> nodeFunnel = (StructNode node, PrimitiveSink into) ->
     {
-      into.putLong(node.getChildCount()).putLong(node.getDescendantCount()).putLong(node.getLeftSiblingKey()).putLong(
-          node.getRightSiblingKey()).putLong(node.getFirstChildKey());
+      if(lastChild != -2L) {
```
JohannesLichtenberger, line 209: Same as above. Maybe it should be a constant.

JohannesLichtenberger:
That's pretty long PR, thanks a lot :+1:

bundles/sirix-core/src/main/java/org/sirix/node/delegates/StructNodeDelegate.java
Raj-Datta-Manohar: Uh oh!! Will change

JohannesLichtenberger:
@all-contributors please add @Raj-Datta-Manohar for code

JohannesLichtenberger:
Thanks for the work. Looking forward to the tests :+1: and maybe further contributions if you like to and time permits :-)

Raj-Datta-Manohar:
Sure:-), I plan to work on the open issues as well.

JohannesLichtenberger:
BTW: do you have some database internals knowledge as you seem to work on Presto?

Raj-Datta-Manohar:
Yes I do, I have also implemented a distributed in-memory database as a project!

JohannesLichtenberger:
Is it your Ph.D. project? :+1:

That's great. I'm thinking about some sort of a simple group commit for the read-write transactions (single writer per resource): https://github.com/sirixdb/sirix/issues/296 and https://github.com/sirixdb/sirix/issues/299 or otherwise version the IndirectPages as well.

The issue is, that SirixDBs log-structure is also the main index, without the need for a WAL, as the main root page (the UberPage) can be swapped atomically.

The data pages are versioned, based on a sliding snapshot algorithm per default, meaning only page fragments are stored and currently a page has pointers to it's former versions. Thus, all nodes/records are stored in a new versionen of the page (a page-fragment), which have been changed in the current version or which would fall out of a sliding window. The offsets into the storage file currently are stored in the data page fragment itself, however they should reside in the PageReferences of the IndirectPages, such that Sirix can fetch the page fragments in parallel if the hardware permits this.

The basic problem I'm currently thinking about is how to avoid write amplification for small read-write transaction commits.

We could also version the IndirectPages, which form the simple trie and reading these fully in-memory (because it would involve up to n-page fragment fetches for each IndirectPage to reconstruct the page in-memory) or we could introduce a form of group commit. A group commit, however would complicate things, such as storing an author + commit message optionally during a commit, which is one of the main benefits of SirixDB, I think. In some sense SirixDB is very similar to Datomic, which also uses a persistent, durable log. However, they accumulate changes and store a log in DynamoDB for instance and they have a log + the real database I think.

I think our system otherwise is very similar to whats described in [1] from Caetano Sauer et al.

[1] http://www.vldb.org/pvldb/vol11/p2249-sauer.pdf

JohannesLichtenberger:
Regarding Presto, I'm also working on Brackit, as Sebastian seems he doesn't have any time anymore: 

http://wwwlgis.informatik.uni-kl.de/cms/dbis/projects/brackit/mission/
http://wwwlgis.informatik.uni-kl.de/cms/fileadmin/publications/2013/Dissertation-Baechle.pdf

and the "fork" with some enhancements regarding JSON and temporal XPath axis: https://github.com/sirixdb/brackit

Raj-Datta-Manohar:
No it was a personal project, I just completed my undergraduation:-)  
I think group commit is not a solution  
I will checkout the Brackit repo!!!

JohannesLichtenberger:
Sounds great :-)

 BTW: it would be great to have some university research around the project again, as it's basically a fork of TreeTank, which is not maintained anymore. Thus, it started as a university project from 2006 - 2013 I think.

PR #303, opened 2020-08-08
Upgrade to Gradle 6.5.1

btmalone:
Ran ./gradlew wrapper --gradle-version 6.5.1 to upgrade the
reference to gradle. Upgraded the Dockerfile to use the correct
version of the Gradle as well.  I was able to build the jar
locally as well as a new Docker image without any issues.

In addition I added a check to verify that you running with the
proper version of Java. When I was first trying to get things
to compile gradle would fail with a cryptic message about plugins
because I was using a earlier version of Java. Hopefully this
will prevent that from happening in the future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,9 @@
 apply from: "${rootDir}/libraries.gradle"
 
+if (JavaVersion.current() != JavaVersion.VERSION_14){
+    throw new GradleException("This build must be run with Java 14")
+}
+
 buildscript {
     repositories {
         maven {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -1,5 +1,9 @@
 apply from: "${rootDir}/libraries.gradle"
 
+if (JavaVersion.current() != JavaVersion.VERSION_14){
```
Raj-Datta-Manohar, line 3: Replace `JavaVersion.VERSION_14` with `targetCompatibilty`

JohannesLichtenberger:
Thanks, looking good :+1:

PR #334, opened 2020-09-27
252 create a kolin cli update command

ManfredNentwig:
Disabled the Test due some Issues wit the update command. The JSON Update works like an insert and the XML Update throws an Exception. They are not really finished jet, I will do that after these Issues are solved.


JohannesLichtenberger:
Can you disable the failing tests? Afterwards I'll merge :-)

JohannesLichtenberger:
I've committed a fix :+1:

PR #340, opened 2020-10-01
Update header.txt

Murrium123:
Use current year


JohannesLichtenberger:
The year is updated accordingly (variable).

Issue #297, opened 2020-07-29
Update REST-API to Vert.x 4 Beta 1


JohannesLichtenberger:
We should update the REST-API (server-rest-api bundle) to Vert.x 4 Beta 1. Some stuff has been deprecated, though.

JohannesLichtenberger:
I think they are putting up a guide how to migrate.

PR #347, opened 2020-10-08
Bumping vertx version to 4.0.0.Beta1

cunhazera:
Closes https://github.com/sirixdb/sirix/issues/297


cunhazera:
@JohannesLichtenberger is there anything wrong with the build?

mosheduminer:
I believe that various vertx APIs have changed between the 3.*.* versions and version 4.0.0, and the code would need to be upgraded to work with the new vertx APIs.

JohannesLichtenberger:
Yes, exactly what @mosheduminer described.

cunhazera:
My bad... I was trying to run without Java 14.

cunhazera:
@JohannesLichtenberger @mosheduminer 
I'm getting this error at the github build: `Caused by: org.gradle.api.GradleException: This build must be run with Java 14`
But locally it fails at all integration tests `./gradlew clean build`
Any idea why?

mosheduminer:
@cunhazera can you clarify what you mean by github build? The travis CI build fails on the integration tests as well, and is using Java 14.

cunhazera:
It don't know much about gradle. Should I really change gradle/wrapper/gradle-wrapper.properties, build.gradle and gradle/wrapper/gradle-wrapper.properties files?

mosheduminer:
It does seem that you introduced changes into build.gradle. I'm not sure why you added these, but it appears that you are targeting java 8, when the project can only be compiled with java 14. But I am not sure that actually causes the problem here.

(I only work on other sirixdb repositories, I probably know less about gradle than you do 😀, Johannes will probably have a better idea)

cunhazera:
Maybe these changes were added by my intellij, idk. Let me remove that and see what happens.

cunhazera:
Can you guys give me some help with this issue?

mosheduminer:
I would guess that there is an authentication issue, which results in an error response, which results in the json errors.

PR #349, opened 2020-10-12
Add max number of nodes and add rightSiblingKey stack for serializing with maxDepth

JohannesLichtenberger:


bundles/sirix-core/src/test/java/org/sirix/service/json/serialize/JsonSerializerTest.java
```diff
@@ -394,6 +394,46 @@ public void testJsonDocumentWithMaxLevel() throws IOException {
...
+      try (final Writer writer = new StringWriter()) {
+        final var serializer = new JsonSerializer.Builder(manager, writer).maxLevel(2).numberOfNodes(5).build();
+        serializer.call();
+
+        final var expected = "{\"foo\":[],\"bar\":{}}";
+        assertEquals(expected, writer.toString());
+      }
+
+      try (final Writer writer = new StringWriter()) {
+        final var serializer = new JsonSerializer.Builder(manager, writer).maxLevel(2).numberOfNodes(6).build();
```
Raj-Datta-Manohar, line 428: This seems redundant, it would be clean if we write another method like
`private String getSerializedString(int maxLevel, int numberOfNodes)`

bundles/sirix-core/src/main/java/org/sirix/axis/visitor/VisitorDescendantAxis.java
```diff
@@ -54,13 +54,13 @@
...
-  private Deque<Long> mRightSiblingKeyStack;
+  private Deque<Long> rightSiblingKeyStack;
 
   /** Optional visitor. */
-  private NodeVisitor mVisitor;
+  private NodeVisitor visitor;
 
   /** Determines if it is the first call. */
-  private boolean mFirst;
+  private boolean first;
```
Raj-Datta-Manohar, line 63: `isFirstCall` would be more appropriate name I guess :-)

Raj-Datta-Manohar (approved):
Looks good to me!

JohannesLichtenberger:
Thanks, for the code review :)

PR #351, opened 2020-10-14
Set storage type depending on some variables and some refactorings

JohannesLichtenberger:


bundles/sirix-core/src/main/java/org/sirix/utils/OS.java
```diff
@@ -4,6 +4,16 @@
 
   public static final boolean IS_64_BIT = is64Bit0();
 
+  private static final String OS = System.getProperty("os.name").toLowerCase();
+
+  private static final boolean IS_LINUX = OS.startsWith("linux");
+
+  private static final boolean IS_MAC = OS.contains("mac");
```
Raj-Datta-Manohar, line 11: Are these needed?

bundles/sirix-core/src/main/java/org/sirix/utils/OS.java
```diff
@@ -4,6 +4,16 @@
...
+
+  private static final boolean IS_WIN = OS.startsWith("win");
+
+  private static final boolean IS_WIN10 = OS.equals("windows 10");
```
Raj-Datta-Manohar, line 15: this too!

bundles/sirix-core/src/main/java/org/sirix/utils/OS.java
Raj-Datta-Manohar, line 11: are these needed?

Raj-Datta-Manohar (approved):

bundles/sirix-core/src/main/java/org/sirix/utils/OS.java
Raj-Datta-Manohar: Everything seems fine

PR #381, opened 2021-02-20
Introduce Dagger as DI framework for the project

### LINKED ISSUE #343: Add a compile time DI framework
We should add a compile time framework (Dagger2?) to make it easier to mock dependencies of a class.

JPDSousa:
Adds Dagger as a dependency to the project. Dagger is used in sirix-core, to create a `DatabaseManager` component.  This component, in this initial stage will simple register dependencies to create `Database` instances. In order to achieve this, a new `LocalDatabaseFactory` concept was introduced, with implementations for Json and XML.

At this point, the component is hidden behind `Databases`, in order to avoid breaking the entrypoint.

The next stages in this track is to extract most of the `Databases` business logic (e.g., sessions map) into separate classes, in order to increase maintainability and testing.

Closes #343

### DIFF
--- a/bundles/sirix-core/src/main/java/org/sirix/access/DatabaseManager.java
+++ b/bundles/sirix-core/src/main/java/org/sirix/access/DatabaseManager.java
@@ -1,0 +1,22 @@
+package org.sirix.access;
+
+import dagger.Component;
+import org.sirix.api.json.JsonResourceManager;
+import org.sirix.api.xml.XmlResourceManager;
+
+import javax.inject.Singleton;
+
+/**
+ * The Dagger component that manages database dependencies. This class is internal and managed by {@link Databases}.
+ *
+ * @author Joao Sousa
+ */
+@Component(modules = DatabaseModule.class)
+@Singleton
+public interface DatabaseManager {
+
+    LocalDatabaseFactory<JsonResourceManager> jsonDatabaseFactory();
+
+    LocalDatabaseFactory<XmlResourceManager> xmlDatabaseFactory();
+
+}
--- a/bundles/sirix-core/src/main/java/org/sirix/access/DatabaseType.java
+++ b/bundles/sirix-core/src/main/java/org/sirix/access/DatabaseType.java
@@ -1,10 +1,6 @@
 package org.sirix.access;
 
-import org.sirix.access.json.JsonResourceStore;
-import org.sirix.access.xml.XmlResourceStore;
 import org.sirix.api.Database;
-import org.sirix.api.NodeReadOnlyTrx;
-import org.sirix.api.NodeTrx;
 import org.sirix.api.ResourceManager;
 import org.sirix.node.SirixDeweyID;
 import org.sirix.node.delegates.NodeDelegate;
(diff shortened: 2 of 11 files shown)


bundles/sirix-core/src/main/java/org/sirix/access/LocalJsonDatabaseFactory.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * A database session factory for Json databases.
+ *
+ * @author Joao Sousa
+ */
+@Singleton
+public class LocalJsonDatabaseFactory implements LocalDatabaseFactory<JsonResourceManager> {
```
JPDSousa, line 16: I know that this sounds dull at this point, but future refactors will definitely make this worth it!

bundles/sirix-core/src/main/java/org/sirix/access/xml/XmlResourceStore.java
```diff
@@ -25,13 +25,6 @@
  */
 public final class XmlResourceStore extends AbstractResourceStore<XmlResourceManager> {
 
-  /**
-   * Default constructor.
-   */
-  public XmlResourceStore() {
-    super(new ConcurrentHashMap<>(), new User("admin", UUID.randomUUID()));
```
JPDSousa, line 32: We're already decoupling this concept of "default user" from the resource store implementation :)

bundles/sirix-core/src/main/java/org/sirix/access/DatabaseManager.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * The Dagger component that manages database dependencies. This class is internal and managed by {@link Databases}.
+ *
+ * @author Joao Sousa
+ */
+@Component(modules = DatabaseModule.class)
+@Singleton
+public interface DatabaseManager {
+
+    LocalDatabaseFactory<JsonResourceManager> jsonDatabaseFactory();
```
JPDSousa, line 18: I left some methods without javadocs, as I'm not sure what is the policy here. Feel free to request further docs if you think it is importante.

JPDSousa:
@JohannesLichtenberger can you give this a look? It's a starting point towards dependency injection and single responsibility principle. I actually have 3 more branches on top of this one from where I can open PRs

JohannesLichtenberger:
Thanks a lot, will review later on :-)

bundles/sirix-core/src/main/java/org/sirix/access/DatabaseType.java
```diff
@@ -23,9 +19,9 @@
   XML("xml") {
     @SuppressWarnings("unchecked")
     @Override
-    public <R extends ResourceManager<? extends NodeReadOnlyTrx, ? extends NodeTrx>, S extends ResourceStore<R>> Database<R> createDatabase(
```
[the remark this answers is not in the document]
JPDSousa: Interesting. I actually though those bounds were implicit (i.e., redundant), hence why I removed them, but apparently, we can have weird things like `? extends Runnable` and the compiler won't complain.

PR #383, opened 2021-02-24
Extract database session management to a separate class

### LINKED ISSUE #382: Extract database session management to an isolated dependency
### Feature idea

At this point, database session management via [Databases](https://github.com/sirixdb/sirix/blob/master/bundles/sirix-core/src/main/java/org/sirix/access/Databases.java#L37), which acts as the API entrypoint for the system. This introduces unnecessary coupling in the class, and also makes testing harder. There is actually a tiny concurrency bug in the way we register new sessions (the `HashSet` in each entry is not handled in a thread-safe manner).

My proposal is for the session management to be extracted to its own class, which we can control via the local databases (each database is responsible for registering and deregistering itself).

JPDSousa:
This patch extracts the database session management into its own class. Also, it adds tests to the session management logic. This new class is maintained by Dagger.

Closes #382

### DIFF
--- a/bundles/sirix-core/src/main/java/org/sirix/access/AbstractLocalDatabase.java
+++ b/bundles/sirix-core/src/main/java/org/sirix/access/AbstractLocalDatabase.java
@@ -8,15 +8,18 @@ import com.google.crypto.tink.JsonKeysetWriter;
 import com.google.crypto.tink.KeysetHandle;
 import com.google.crypto.tink.streamingaead.StreamingAeadKeyTemplates;
 import org.sirix.access.trx.TransactionManagerImpl;
-import org.sirix.api.*;
+import org.sirix.api.Database;
+import org.sirix.api.NodeReadOnlyTrx;
+import org.sirix.api.NodeTrx;
+import org.sirix.api.ResourceManager;
+import org.sirix.api.Transaction;
+import org.sirix.api.TransactionManager;
 import org.sirix.cache.BufferManager;
 import org.sirix.cache.BufferManagerImpl;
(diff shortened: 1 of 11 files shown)


bundles/sirix-core/src/main/java/org/sirix/access/AbstractLocalDatabase.java
```diff
@@ -65,16 +67,29 @@
...
    * Constructor.
-   *
-   * @param dbConfig {@link ResourceConfiguration} reference to configure the {@link Database}
+   *  @param dbConfig {@link ResourceConfiguration} reference to configure the {@link Database}
+   * @param sessions
+   * @param resourceStore
    */
-  public AbstractLocalDatabase(final DatabaseConfiguration dbConfig) {
+  public AbstractLocalDatabase(final DatabaseConfiguration dbConfig, final DatabaseSessionPool sessions, final ResourceStore<T> resourceStore) {
+
```
JohannesLichtenberger, line 95: Only one minor thing, can we skip the blank line after a constructor or method declaration? Thanks a lot for all the work :+1:

JohannesLichtenberger: I should provide my most recent IntelliJ formatter, I think, such that we can simply auto-format :-)

JPDSousa: I'm using IntelliJ as well, but I installed a plugin that imports the formatter from Eclipse. Apparently it didn't work as well as I thought 🤔

PR #387, opened 2021-02-27
Remove cyclic dependency between Database and ResourceStore

JPDSousa:
This commit removes the cyclic dependency between `Database` and `ResourceStore´.


JPDSousa:
I can try to fix the code factor problem, but that will introduce a lot of noise to this PR :(

JohannesLichtenberger:
> I can try to fix the code factor problem, but that will introduce a lot of noise to this PR :(

We have to resolve the conflicts, but regarding the code factor problem, I think that's something we can resolve later on, if we like to...

JPDSousa:
> We have to resolve the conflicts

Done @JohannesLichtenberger ✅

PR #388, opened 2021-02-28
Extract resource write lock management to its own class

JPDSousa:
This patch extracts write lock management on `ResourceManager` to its own class. This class is now being registered in the central `DatabaseManager` component.

### DIFF
--- a/bundles/sirix-core/src/main/java/org/sirix/access/WriteLocksRegistry.java
+++ b/bundles/sirix-core/src/main/java/org/sirix/access/WriteLocksRegistry.java
@@ -1,0 +1,65 @@
+package org.sirix.access;
+
+import org.sirix.api.ResourceManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.concurrent.ThreadSafe;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * A registry for write locks used for {@link ResourceManager resource managers}.
+ *
+ * <p>Each {@link ResourceManager}, identified by its {@link Path resource path}, will be
+ * assigned a unique write lock.
+ *
+ * @author Joao Sousa
+ */
+@Singleton
+@ThreadSafe
+public class WriteLocksRegistry {
+
+    /**
+     * Logger for {@link WriteLocksRegistry}.
+     */
+    private static final Logger logger = LoggerFactory.getLogger(WriteLocksRegistry.class);
+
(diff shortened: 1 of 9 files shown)


bundles/sirix-core/src/main/java/org/sirix/access/WriteLocksRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * <p>This method will create a new write lock if no lock exists for the provided resource.
+     *
+     * @param resourcePath The path that identifies the resource.
+     * @return The lock to be used to perform write operations to the given resource.
+     */
+    public Lock getWriteLock(final Path resourcePath) {
+        logger.trace("Fetching log for resource with path {}", resourcePath);
+
+        return this.locks.computeIfAbsent(resourcePath, res -> new ReentrantLock());
```
JPDSousa, line 52: @JohannesLichtenberger This patch raised some concerns (note that despite the refactor, this was already the current behavior of the system):
* The fact that we use reentrant java locks to serialize write transactions is a bit weird. We can create a `Trx1` with `ThreadA` (locks the transaction to `ThreadA`) and then create another `Trx2` with `ThreadA` without committing `Trx1`. Since the reentrant lock allows the same thread to acquire the lock multiple times, there is nothing (that I can see) preventing this.

JohannesLichtenberger: Yes, that's indeed a bug.

JohannesLichtenberger: It's too late right now, but I couldn't find another Lock implementation despite reentrant lock implementations. Will have to research tomorrow ;-)

JohannesLichtenberger: Can you resolve the conflicts with your other PRs?

JPDSousa: > Can you resolve the conflicts with your other PRs?

Will do it today if I can find the time 👍 

In regards to the bug, I think I have a solution: we can have our own `ResourceLock`, and instead of binding it to a thread, we bind it to a transaction id, all in a thread safer way. What do you think?

JohannesLichtenberger: I'm not sure where the write lock has been used exactly. Is it the lock, that only one read-write trx can be opened on a resource? We then should probably bind it to a resource name.

JPDSousa: > Is it the lock, that only one read-write trx can be opened on a resource? We then should probably bind it to a resource name.

That's the one. The lock is already bound to the resource (i.e., we have one lock instance per resource ~ that's the whole goal of the registry). Sorry if it wasn't clear, but my initial example (as well as this whole thread) was based on transactions affecting the same resource.

PR #413, opened 2021-06-04
Update to JDK 16

mosheduminer:
This first requires a new release of brackit, for JDK 16. @JohannesLichtenberger please review.


JohannesLichtenberger:
Can you make a OR for Brackit, too? ;-)

mosheduminer:
Here https://github.com/sirixdb/brackit/pull/11 🙂.

mosheduminer:
It seems like the brackit update wasn't published? Do we need a version bump?

JohannesLichtenberger:
There are sadly other changes to consider:

org.sirix.diff.algorithm.FMSETest > testAllEigth FAILED
    java.lang.invoke.WrongMethodTypeException: cannot convert MethodHandle(VarHandle,MemorySegment,long,int)void to (VarHandle,MemoryAddress,int)void
        at java.base/java.lang.invoke.MethodHandle.asTypeUncached(MethodHandle.java:884)
        at java.base/java.lang.invoke.MethodHandle.asType(MethodHandle.java:869)
        at org.sirix.io.memorymapped.MMFileWriter.write(MMFileWriter.java:188)

I guess something else changed in the foreign memory API.

mosheduminer:
Oh, I seemed to have missed that. I'll look into it.

mosheduminer:
It is very strange, because the tests pass locally (except two json shredder tests, which seems to have encoding issues, maybe windows related?).

JohannesLichtenberger:
Maybe this helps: https://openjdk.java.net/jeps/393

Something clearly changed with method handles, regarding the exception.

Locally you run Windows, right? On Windows the file backend is used instead of the memory mapped backend ;-)

mosheduminer:
> Locally you run Windows, right? On Windows the file backend is used instead of the memory mapped backend ;-)

Oh, of course 😅. I completely forgot. I'll try to fix the changes.

PR #429, opened 2021-08-28
Extract common transaction logic into AbstractNodeTrxImpl

JPDSousa:


JPDSousa:
@JohannesLichtenberger this is 90% complete. I need to fix some stuff (probably tests will break), but at least the gist of it is there. This is still code that was extracted from #410 (I excluded everything that was async commit related from that PR)

JohannesLichtenberger:
Thanks a lot... I had a quick look and of course it makes a lot of sense :+1:

JPDSousa:
It should be good to go now @JohannesLichtenberger 👍

JohannesLichtenberger:
Yes, thanks a lot.

Issue #432, opened 2021-09-16
Update to Java 17


JohannesLichtenberger:
We should update the project to Java 17.

pavluchenko:
Hi, I would like to look on this. Can you please assign to me this issue?

JohannesLichtenberger:
@pavluchenko seems you won't have tine to work on the issue?

Rishavgupta12345:
i world like to work on this

JohannesLichtenberger:
Sure, give it a try :)

PR #439, opened 2021-10-24
Java17

artwo:
# Upgrade Java 17
This PR upgrades the version of Java from 16 to 17 and it is meant to solve the issue https://github.com/sirixdb/sirix/issues/432.

Changes:
- Set JDK 17
- Set Kotlin compile target to JVM 16 (JVM 17 support is still not available)
- Update Gradle for Java 17 compatibility
- Update GH Actions and Docker files
- Update testing dependencies for compatibility changes
- MMFileReader and MMFileWriter updated for changes in the JDK API
- Update Brackit library

### Note:
~~Project is building but tests are not working, this is due to incompatibility with the project https://github.com/sirixdb/brackit. The brackit lib was compiled with preview features that are unsupported in Java 17. Please create a new issue in the brackit project with #Hacktoberfest and I would be happy to work on it.~~
With the update of the Brackit lib the tests and the build are working correctly, but the new Brackit lib version has to be uploaded to the maven repository.


artwo:
So with the brackit updated and available, the build and the checks should work now. Could you please run them again @JohannesLichtenberger ?

mosheduminer:
I retriggered the workflow.

artwo:
Thanks @mosheduminer . I see that for some reason it is trying to use Java 16 in the build with the GH action, but that was updated to 17, so I don't understand why.

artwo:
Which file is it used for the build GH action? It seems to not be using .github/workflows/gradle.yml

mosheduminer:
> Which file is it used for the build GH action? It seems to not be using .github/workflows/gradle.yml

I'm wondering the same thing. I can't figure out where that action is coming from.

mosheduminer:
@artwo looks like the workflow file on master is more extensive than the version on your branch. Maybe try merging in the changes from master?

artwo:
@mosheduminer indeed I have just noticed some changes in the build job in master since I started this PR. Thanks!

artwo:
@mosheduminer the changes are there, could you please have a look again?

artwo:
It seems to be failing now due to a mis-configuration of the new sonar task:
```FAILURE: Build failed with an exception.
> Task :sonarqube FAILED

:sonarqube (Thread[Execution worker for ':',5,main]) completed. Took 12.352 secs.
* What went wrong:

Execution failed for task ':sonarqube'.
> You're not authorized to run analysis. No sonar.login or SONAR_TOKEN env variable was set
```

mosheduminer:
I think the token is set, but I believe that we are using the community edition of Sonar, and that it only allows for analysis of the main branch, therefore failing on your branch/fork/PR. I suggest we merge this and fix this in the future. What do you think @JohannesLichtenberger?

artwo:
There could be a dedicated GH action that runs Sonar on the main branch, then disable Sonar everywhere else. But this is just a suggestion for an improvement in another PR.

artwo:
Here is the improvement I mentioned before: https://github.com/sirixdb/sirix/pull/442

JohannesLichtenberger:
Thanks a lot, would be amazing of you contribute beyond the Hacktoberfest ;)

artwo:
It's a pleasure! Thank you for accepting. I will try to send more PRs if the time allows me to ;)

Issue #440, opened 2021-10-25
Fix Sonar issues


JohannesLichtenberger:
https://sonarcloud.io/summary/overall?id=sirixdb_sirix

gvdutra:
Hello @JohannesLichtenberger Can I work on this? I am available to fix a few sonar issues. thanks.

JohannesLichtenberger:
Of course :-)

slashgk:
Hi @JohannesLichtenberger Raised a PR: https://github.com/sirixdb/sirix/pull/545

slashgk:
Hi @JohannesLichtenberger 
Raised a PR: https://github.com/sirixdb/sirix/pull/667

persy97:
Hi @JohannesLichtenberger
Raised a PR: https://github.com/sirixdb/sirix/pull/669

JohannesLichtenberger:
@persy97 that's not enough for a meaningful contribution, sorry

persy97:
@JohannesLichtenberger Pushed some more sonar fixes in the same PR. Please take a look.

JohannesLichtenberger:
Thanks, but I think a cast is missing :)

persy97:
Can you trigger the build again?

Naman0133:
Hi @JohannesLichtenberger raised a PR: [#671](https://github.com/sirixdb/sirix/pull/671)

Salonimalhotra:
Hi @JohannesLichtenberger raised a PR: https://github.com/sirixdb/sirix/pull/674
can someone from maintainer approve?

smirti-barnwal:
Hi @JohannesLichtenberger raised a PR: https://github.com/sirixdb/sirix/pull/676
can someone from maintainer approve?

PR #451, opened 2021-12-13
Fix Sonar issues #440

gvdutra:
Still, a long list to be solved, I will be doing it as I have time.

### DIFF
--- a/bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractNodeHashing.java
+++ b/bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractNodeHashing.java
@@ -158,7 +158,7 @@ public abstract class AbstractNodeHashing<N extends ImmutableNode> {
       cursorToRoot = pageTrx.prepareRecordForModification(getCurrentNode().getNodeKey(), IndexType.DOCUMENT, -1);
       hashCodeForParent = getCurrentNode().computeHash().add(hashCodeForParent.multiply(PRIME));
       // Caring about attributes and namespaces if node is an element.
-      if (cursorToRoot.getPathKind() == NodeKind.ELEMENT) {
+      if (cursorToRoot.getKind() == NodeKind.ELEMENT) {
         final ElementNode currentElement = (ElementNode) cursorToRoot;
         // setting the attributes and namespaces
         final int attCount = ((ElementNode) cursorToRoot).getAttributeCount();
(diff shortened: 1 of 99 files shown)


JohannesLichtenberger:
Can you fix the failing tests?

gvdutra:
yes for sure, so strange I ran locally and it was all good. lemme check it.

JohannesLichtenberger:
Can you revert this path kind change? Meaning the `getNodeKind()` should be right.

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractNodeHashing.java
```diff
@@ -158,7 +158,7 @@ private void postorderAdd() {
       cursorToRoot = pageTrx.prepareRecordForModification(getCurrentNode().getNodeKey(), IndexType.DOCUMENT, -1);
       hashCodeForParent = getCurrentNode().computeHash().add(hashCodeForParent.multiply(PRIME));
       // Caring about attributes and namespaces if node is an element.
-      if (cursorToRoot.getKind() == NodeKind.ELEMENT) {
+      if (cursorToRoot.getPathKind() == NodeKind.ELEMENT) {
```
JohannesLichtenberger, line 161: this should be the `getKind()`-method, but somewhere I think you changed something between `getKind()` and `getPathKind()`. But over here it checks the type of the node itself, not the path type/kind.

JohannesLichtenberger: Accordingly all the other changes in this commit :)

gvdutra: ok, I will revert the getKind and getPathKind that I changed on main/java/org/sirix/index/path/summary/PathNode.java for now, but, just for you know sonar is complaining as a critical bug. https://sonarcloud.io/project/issues?id=sirixdb_sirix&open=AXy4_zUSQgQXZ6uows07&resolved=false&severities=CRITICAL&types=BUG

gvdutra:
Done :) if anything needs to be changed please let me know, after this gets merged, I will continue working on more sonar issues.

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractNodeHashing.java
gvdutra: Done :) if anything needs to be changed please let me know, after this gets merged, I will continue working on more sonar issues.

JohannesLichtenberger: getKind() should return NodeKind.PATH as that's the type of the PathNode. The `getPathKind()`-method should return which type of a path node the current node denotes (for JSON: array or object fields, for XML: attribute or element).

That is a path node is a node of a path summary, which simply is a "small" tree, which denotes all paths in a resource :-) Hope that makes sense and helps in getting the context. Did you already push your revert?

gvdutra: Oh I see, so we might need to add the @ignore tag for sonar on this class and/or might be useful to add a comment on it, as Sonar thinks this is an issue. yes, changes on PathNode class are reverted! :)

JohannesLichtenberger: Yes, I've closed the issue on Sonar with a short description. We have to be careful about false positives.

I can't find the reverted changes. `AbstractNodeHashing` for instance still uses `getPathKind()` as far as I can see.

gvdutra: it got messy I will commit again in another PR.

JohannesLichtenberger: Thanks and looking forward to your contributions :-)

PR #462, opened 2022-01-09
fixed some sonar issues #440

gvdutra:


bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
```diff
@@ -384,9 +384,6 @@ public synchronized W beginNodeTrx(final @Nonnegative int maxNodeCount, final @N
       final @Nonnull TimeUnit timeUnit, final @Nonnull AfterCommitState afterCommitState) {
     // Checks.
     assertAccess(getMostRecentRevisionNumber());
-    if (maxNodeCount < 0 || maxTime < 0) {
```
JohannesLichtenberger, line 387: Hm, I'm not sure if we should solely rely on the annotations in this case. What do you think, @JPDSousa?

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
```diff
@@ -519,11 +514,10 @@ public void setNodePageWriteTransaction(final @Nonnegative long transactionID, @
    * @throws SirixIOException if an I/O error occurs
    */
   @Override
-  public void closeNodePageWriteTransaction(final @Nonnegative long transactionID) {
+  public void closeNodePageWriteTransaction(final @Nonnegative Long transactionID) {
```
JohannesLichtenberger, line 517: Why is it better to use a boxed version over here?

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
```diff
@@ -647,16 +641,15 @@ public PageReadOnlyTrx beginPageReadOnlyTrx(final @Nonnegative int revision) {
     assertAccess(revision);
 
     final long currentPageTrxID = pageTrxIDCounter.incrementAndGet();
-    final NodePageReadOnlyTrx pageReadTrx =
-        new NodePageReadOnlyTrx(currentPageTrxID, this, lastCommittedUberPage.get(), revision, storage.createReader(),
-            null, bufferManager, new RevisionRootPageReader());
+    try (NodePageReadOnlyTrx pageReadTrx = new NodePageReadOnlyTrx(currentPageTrxID, this, lastCommittedUberPage.get(), revision, storage.createReader(),
```
JohannesLichtenberger, line 644: This doesn't work as the page trx shouldn't be closed immediately.

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
JPDSousa: @gvdutra Note that if the issue is the auto-boxing happening in `nodePageTrxMap.remove(transactionID)`, I would rather turn `nodePageTrxMap` into a [Long2ObjectMap](https://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/longs/Long2ObjectMap.html) as native types are more performant (e.g., there's no extra pointers/space in the heap memory) and we prevent `null` out of the box.

bundles/sirix-core/src/main/java/org/sirix/service/xml/shredder/WikipediaImport.java
```diff
@@ -276,31 +276,33 @@ private void updateShredder() throws SirixException, IOException, XMLStreamExcep
...
+        rtx.moveToFirstChild();
+        final long nodeKey = mWtx.getNodeKey();
+        try (final FMSE fmse = FMSE.createInstance(new DefaultNodeComparisonFactory())) {
+          fmse.diff(mWtx, rtx);
+        }
+        mWtx.moveTo(nodeKey);
+        rtx.close();
+        resourceManager.close();
+      }
+      db.close();
```
JPDSousa, line 304: Shouldn't we remove all these `.close()` as the resources are automatically closed by `try`?

bundles/sirix-core/src/main/java/org/sirix/service/xml/shredder/WikipediaImport.java
```diff
@@ -572,7 +574,7 @@ public static void main(final String[] args) throws SirixException {
         eventFactory.createStartElement(new QName(NSP_URI, "text", XMLConstants.DEFAULT_NS_PREFIX), null, null);
 
     // Create list.
-    final List<StartElement> list = new LinkedList<StartElement>();
+    final List<StartElement> list = new LinkedList<>();
```
JPDSousa, line 577: While you're at it: can't this be changed to an `ArrayList<>(5)` or an `ImmutableList.of(...)`, so that we get a perfectly sized array, rather than a scattered node-backed list (which is less read-performant)?

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
JohannesLichtenberger: @JPDSousa I'd argue that we still need the precondition checks to fail fast: https://stackoverflow.com/questions/13484202/how-to-use-nullable-and-nonnull-annotations-more-effectively

For me it's just a documentation and I've not used static type checkers, but even if I think it's necessary to still check explicitly, don't you think?

bundles/sirix-core/src/main/java/org/sirix/service/xml/shredder/WikipediaImport.java
JohannesLichtenberger: True, after the change it's implicitly closed, even probably before (I'd have to check later on after work, what the FMSE close() method does.

gvdutra:
thanks for all the comments! I will work on it today.... :)

bundles/sirix-core/src/main/java/org/sirix/service/xml/shredder/WikipediaImport.java
JohannesLichtenberger: Now you can simply use `List.of(...)`.

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
JPDSousa: I see your point, and the fail-fast argument is a really strong one. I missed the detail that if we don't do this check, we will catch the invalid value too late (e.g., nodeTrxIDCounter.incrementAndGet() would have already ran). 👍 

However, let me re-emphasize that these annotations should live in the base method (declared in the interface), as that's where the "official" contract definition is. Also, I would then argue that declaring them here is redundant, as this class is not even meant to be API-visible.

bundles/sirix-core/src/main/java/org/sirix/service/xml/shredder/WikipediaImport.java
JPDSousa: Thanks @JohannesLichtenberger I guess I'm becoming outdated in the Java world 😅

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
JohannesLichtenberger: I see your point, you're right :-)

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
gvdutra: As nodePageTrxMap is ConcurrentMap, I was afraid to change it to Long2ObjectMap, I couldn't find the artifact imported, also as I don't know much about the project, I am not sure if Long2ObjectMap would impact in somewhere else, or even if it is threadsafe!

JohannesLichtenberger:
Can you add the precondition check in `AbstractResourceManager` again: `if (maxNodeCount < 0 || maxTime < 0) {`. Furthermore, the annotations should be in the interface instead of the implementing class as discussed with @JPDSousa :-)

gvdutra:
A really stupid question, haha... 

Should I leave the preconditions on the implementation as well? I was looking for it on internet, and some ppl says it is good to leave it, and some others are saying it is just repetition. IMO, it is ok both scenarios. What pattern do you want to apply?

Also, I am adding the precondition on the interface for the other implemented methods as well.

JohannesLichtenberger:
I think the precondition checks alone should reside in the implementing class or default methods in the interfaces, the Annotations in the interfaces, I guess :-)

gvdutra:
Guys, Looks like annotation only in the interface doesn't work.  So, I kept it on the implementation as well.

https://stackoverflow.com/questions/28894461/do-i-need-to-nonnull-again-at-the-implementation

Also, another point is. findbugs is deprecated and is not maintained anymore, people are using this  https://github.com/spotbugs/spotbugs

If you all agree I can create an issue to replace and work on it as well. Just lemme know.

Please if I am missing something say it :). Thanks.

bundles/sirix-core/src/main/java/org/sirix/access/trx/node/AbstractResourceManager.java
JohannesLichtenberger: I'm also not sure, but it has to be threadsafe. Maybe just leave it as of now?

bundles/sirix-core/src/main/java/org/sirix/api/ResourceManager.java
```diff
@@ -164,7 +164,7 @@ default R beginNodeReadOnlyTrx() {
    * @throws SirixThreadedException   if the thread is interrupted
    * @throws SirixUsageException      if the number of read-transactions is exceeded for a defined time
    */
-  R beginNodeReadOnlyTrx(@Nonnull Instant pointInTime);
+  R beginNodeReadOnlyTrx(final @Nonnull Instant pointInTime);
```
JohannesLichtenberger, line 167: I know that I sometimes specified final for interface method params. However, I'm not sure if that's even a good practice. Usually I don't do it anymore, as it has no effect, but maybe it's a kind of documentation at least. Even though I tend to remove them. What do you think and @JPDSousa?

JPDSousa: `final` just prevents parameter reassignment, which for interfaces has no effects. I think you can even "override" final in your implementation by simply not including it there. Therefore I think final only makes sense in methods which are implemented.

Also note that parameter reassignment is a bad practice [(source)](https://stackoverflow.com/questions/13507025/avoid-reassigning-parameters). `final` is just a way of making it explicit.

gvdutra: you are right Joao, my bad. I will remove it.

PR #468, opened 2022-01-17
removed final keyword from abstract methods params #440

gvdutra:
I also fixed a javadoc param error.


JohannesLichtenberger:
Thanks :-)

PR #474, opened 2022-02-10
Set up git pre commit hook.

omegaspard:


omegaspard:
@JohannesLichtenberger 
I have some questions here:
- I'm using the spotless task that already had the `licenseHeaderFile` instruction. The issue with it is that it would break the java class because the content of the LICENSE file isn't commented. I commented it out for now.
- In the initial eclipse formater file I tried to use in the `eclipse` instruction of the spotlessJava task, multiple profiles were defined. Apparently it shouldn't contains multiple profile so I created one file for each and used only one.
 image
- To make the pre commit hooks works, I put the sh files into the .github/hooks/ because the default folder it should be in is .git and it's not versioned. The issue is that it force the developer to execute `git config --local core.hooksPath .github/hooks` in order to have the hooks. I know that you can set up remote hooks at the organization level on git.

What should I do with all that ?

JohannesLichtenberger:
Regarding 1. and 2. seems fine. I'll finally push my IntelliJ formatter tomorrw as promised to do already days ago.

Regarding 3. I'm not sure where. Maybe you know  @mosheduminer or @JPDSousa?

mosheduminer:
In my experience, a git commit hook needs to be installed by the user, there's no way to force it to be automatically enforced when someone `git clone`s.
Of course, I may be wrong, however.

JohannesLichtenberger:
As far as I can see the `post-receive` hook should be the right hook.

omegaspard:
> Actually I think we need pre commit hooks for us and also a `pre-receive` hook which does auto-formatting, too and compares the result with the pushed changes and maybe denies the push. But I'm not sure if that would discourage people to even contribute.

What you describe is what I have in my current job. It's annoying sometimes but it does work well to maintain a uniformed code style.

You can refers to this documentation.

https://docs.github.com/en/enterprise-server@3.0/admin/policies/enforcing-policy-with-pre-receive-hooks/managing-pre-receive-hooks-on-the-github-enterprise-server-appliance#configure-pre-receive-hooks-for-an-organization


> In my experience, a git commit hook needs to be installed by the user, there's no way to force it to be automatically enforced when someone `git clone`s. Of course, I may be wrong, however.

We could reduce that to just asking the developer to change the hooks folder for this particular project. Or setup a sim link between the .github/hooks folder and its actual hooks folder.

JohannesLichtenberger:
I think it's an enterprise feature on Github, so can't find the Hooks menu item in the organization settings.

Maybe it's easiest to simply auto-format the code in a save action in IntelliJ or Eclipse for instance as I'm usually doing. But for sure nothing is enforced, of course. I think this organizational feature however is nice, but currently we're only spending money on SirixDB, so it's maybe something if we'll be able to monetarize the project sometime.

JohannesLichtenberger:
I've added the new formatter (SirixDB.xml) in the root folder.

JPDSousa:
> In my experience, a git commit hook needs to be installed by the user, there's no way to force it to be automatically enforced when someone `git clone`s. Of course, I may be wrong, however.

I second this. I know that we can set a lint/style-check job via github actions. That will prevents PRs to be merge with stylecheck issues. However, it will always fallback to the user to deal with it, either with a manually installed hook, or just doing manual lint.

IMO we should be careful with forcing lint, as it might demotivate contributors (i.e., it introduces additional friction).

mosheduminer:
I wonder if it is possible to have a gh action that will reformat (not just check) the code? I also wonder if that's a good idea. Maybe an action on master (for when PRs are merged)?

JohannesLichtenberger:
I did a quick search and I think it should be possible.

JPDSousa:
> I wonder if it is possible to have a gh action that will reformat (not just check) the code? I also wonder if that's a good idea. Maybe an action on master (for when PRs are merged)? 

I'm sure that's possible, but we need to be careful with the fact that developers will have an extra commit on their PRs, as these actions usually auto-commit. Hence, the development flow becomes more error prone, as we can to push+pull.

I would suggest that we should maybe follow the master approach (I.e. post-merge) to mitigate this issue, but I'm not experienced with this flow, and it's not obvious to me what the respective pros/cons would be.

omegaspard:
Sry I put this on hold..

What do we do with it then ? It seems that people aren't convinced of this feature.

@JPDSousa 

> I'm sure that's possible, but we need to be careful with the fact that developers will have an extra commit on their PRs, as these actions usually auto-commit. 

I could put a `commit --amend --no-edit` ? It would be transparent for the devs.

> I would suggest that we should maybe follow the master approach (I.e. post-merge) to mitigate this issue, but I'm not experienced with this flow, and it's not obvious to me what the respective pros/cons would be.

An alternative would be to block the push if the code isn't properly formatted. 

@JohannesLichtenberger 

On the technical side I don't find any tool to format the code based on the Intelij IDEA code style xml file. The only thing I found doing this is the actual sh launcher of intelij IDEA but it doesn't fit what we want since it force to have Intelij IDEA installed.

On the other end we can call eclipse code style xml file and some other stuff using spotless through gradlew.

JPDSousa:
Hey @omegaspard 

No worries with the timing! Feel free to take your time!

> > I'm sure that's possible, but we need to be careful with the fact that developers will have an extra commit on their PRs, as these actions usually auto-commit.
> 
> I could put a `commit --amend --no-edit` ? It would be transparent for the devs.

There is one caveat with your approach. `git commit --amend` actually creates a new commit and replaces it by the old one ([source](https://careerkarma.com/blog/git-commit-amend/#:~:text=The%20old%20commit%20is%20replaced%20with%20a%20new%20commit)). This means that developers would actually get something like this:
```
Your branch and 'origin/master' have diverged,
and have 1 and 1 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
nothing to commit, working tree clean
```
Which would be more confusing than simply adding a new commit.


> 
> > I would suggest that we should maybe follow the master approach (I.e. post-merge) to mitigate this issue, but I'm not experienced with this flow, and it's not obvious to me what the respective pros/cons would be.
> 
> An alternative would be to block the push if the code isn't properly formatted.

This looks like a viable approach to me 👍 

# Action points

I think that we have 2 possible approaches here

## Plan A
1. Pre-commit hook that autoformats the code
2. New github action that checks code formatting, for users that don't setup the pre-commit hook
3. The github action can possibly be set to warning to not demotivate developers from contributing, even when they don't feel like formatting the code

## Plan B
1. Pre-push hook that asserts no unformatted code is pushed to the remote repo
2. Possibly follow the same steps (2. and 3.) as Plan A, for the same reasons

## Personal opinion
I intentionally set the pre-commit hook approach as plan A, due to the following pros:
* AFAIK it's the most common approach
* It relieves developers from a lot of auto-formatting
* It has exactly the same setup cost as Plan B
* It tolerates developers who don't like to format code

JPDSousa (approved):

JohannesLichtenberger:
I'm also for Plan A, but it would be awesome if we could use the IntelliJ formatter :)

Issue #467, opened 2022-01-17
Replace JSR-305 for checkerframework


gvdutra:
### Feature idea

http://findbugs.sourceforge.net/index.html has not been maintained anymore it stops on java 8. People are now using spotbugs which is a fork and continuation of findbugs. https://spotbugs.github.io/

The idea here is to replace findbugs artifactory for spotbugs.

gvdutra:
@JohannesLichtenberger can you please assign it to me? Thank you.

JohannesLichtenberger:
Sure, thanks a lot for your contributions :+1:

gvdutra:
Yes! I will do it, sorry I am a bit busy.... :)

JohannesLichtenberger:
No pressure, just wanted to ask ;)

gvdutra:
@JohannesLichtenberger sirix project is using JSR-305 for the annotations, not the "entire" findbugs, so the spotbugs is not ideal as still uses JSR-305.

JSR-305 is in dormant status even at spotbugs it says that to keep using JSR-305 from findbugs. , so that I was searching for a better solution, and I found it! there is a project called https://checkerframework.org/manual/ that is sort of slf4j for logging framework, whereas the "logging framework" here is all those validation in form of annotation. So that, I decided to move to checkerframework, I think this is the better solution IMO. 

Lemme know what you think, and sorry for the delay, I was trying to find a better solution for it.

PR #476, opened 2022-02-14
Replace JSR-305 for checkerframework on sirix-core #467

gvdutra:


gvdutra:
As I am having some issues with replacing it completely, I am breaking down the PR for every sirix sub-project to be on the "safe" side.

JPDSousa:
@gvdutra I think this is one of those PRs where a description is of upmost important? Why are we replacing javax by a third library? Can you include some links that point out to this being a good practice?

From a code standpoint, this looks awesome as usual 👌

gvdutra:
Hi Joao! Oops ... sorry, I thought as it was linked with the issue it would be ok to leave without a description! I will put it from now on!

I was initially about to replace findbugs, which is dead project to spotbugs. However, I found out that Sirix is only using it because of JSR-305, so that searching more I found out that JSR-305 is also dead! After searching and searching, I found out this third party project, that is like a interface with implementation for all those static validation in the market.  https://checkerframework.org/manual/#faq-jsr-305

https://jcp.org/en/jsr/detail?id=305 -> Status: [Dormant](https://jcp.org/en/introduction/glossary#status)

https://github.com/sirixdb/sirix/issues/467#issuecomment-1038628300


Please, if you think this is not a good idea feel free to decline the PR, no worries at all!

Many thanks!

JohannesLichtenberger:
I think it's okay to switch and we are then probably able to really check the code at compile time as far as I understand!?

What do you think @JPDSousa? If you have no objections I'd otherwise merge the changes!?

@gvdutra thanks for all your work. If you'd like to work on features at some point or other refactoring stuff keep up the great work and let me know :+1:

JPDSousa (approved):

JPDSousa:
> Hi Joao! Oops ... sorry, I thought as it was linked with the issue it would be ok to leave without a description! I will put it from now on!
> 
> I was initially about to replace findbugs, which is dead project to spotbugs. However, I found out that Sirix is only using it because of JSR-305, so that searching more I found out that JSR-305 is also dead! After searching and searching, I found out this third party project, that is like a interface with implementation for all those static validation in the market. https://checkerframework.org/manual/#faq-jsr-305
> 
> https://jcp.org/en/jsr/detail?id=305 -> Status: [Dormant](https://jcp.org/en/introduction/glossary#status)
> 
> [#467 (comment)](https://github.com/sirixdb/sirix/issues/467#issuecomment-1038628300)
> 
> Please, if you think this is not a good idea feel free to decline the PR, no worries at all!
> 
> Many thanks!

Sorry @gvdutra I was reviewing this on my phone, and I missed the issue link on the commit title. The usual convention is to include a keyword and the issue id in the PR description (e.g., `Closes #467`). You can read more about it [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

Nevertheless my goal was more towards us making data driven decisions, but looking both at your comment (to which I'm replying) as well as the original issue, I see you've made an investigation and included links for us to understand the reasons behind the PR 🏆 

Once again, sorry for the confusion, and great work!!

PR #499, opened 2022-06-02
Fix grale build warnings throughout the codebase

AlvinKuruvilla:
This commit fixes some gradle build warnings. There is still a lot more to do, but this is a decent dent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ out/
 *.project
 *.classpath
 
+# Visual Studio Code files 
+settings.json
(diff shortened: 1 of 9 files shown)


AlvinKuruvilla:
Not sure why the python client checks fail

JohannesLichtenberger:
Currently they also fail on the master branch. I'm also not sure why the connection fails

AlvinKuruvilla:
I guess just force push for now. We probably also want to do that for the docker plugin updates

PR #529, opened 2022-09-08
Fixed typo in README.md

Philipp-p:
It appears that "ame configuration file" should be "same configuration file" as I seems to refer to the configuration file in the line above


JohannesLichtenberger:
Thanks :-)

PR #539, opened 2022-10-16
Make working with ByteBuffer more safe

ikorennoy:
I noticed that it is possible to make working with ByteBuffer safer.

Imagine a situation where you read fewer bytes than expected, then simply calling `position(0)` and trying to retrieve the read data will probably result in getting incorrect data because that buffer's **limit** will still be the buffer **size.**

However, if you use `flip()`, the buffer **limit** is set to its current **position** (equal to the number of bytes read) and **position** to 0. So, an exception will occur if you try to get more bytes from the buffer than were read.


JohannesLichtenberger:
Thanks a lot :-)

PR #543, opened 2022-10-27
ci: Cache gradle dependencies

jidicula:
Uses actions/cache to cache Gradle dependencies to speed up CI runs. Also stops the Gradle daemon at the end of the job so the cache can be built in the post-workflow steps.

See https://github.com/actions/cache/blob/main/examples.md#java---gradle

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -25,10 +25,20 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '19'
+      - uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


jidicula:
This is also a Hacktoberfest PR - if it passes muster, could you add the `hacktoberfest-accepted` label to this PR?

JohannesLichtenberger:
The build currently fails

jidicula:
> The build currently fails

Missed the `./` - should be fine now 🤞

Issue #541, opened 2022-10-22
Adapt the keycloak startup script


JohannesLichtenberger:
We first have to check if an admin user already exists in this script: https://github.com/sirixdb/sirix/blob/master/bundles/sirix-rest-api/src/test/resources/create-sirix-users.sh

Aditya31398:
Hi @JohannesLichtenberger, I am a first timer. Can I work on this?

JohannesLichtenberger:
Sure :-)

Aditya31398:
Hi @JohannesLichtenberger, can you brief over the issue and let me what changes are required?

JohannesLichtenberger:
The user should only be added if the user hasn't been added before, thus, first check if user exists, if not create the user

JohannesLichtenberger:
@Aditya31398, do you like and have time to work on the issue?

Aditya31398:
@JohannesLichtenberger, sorry for the delay. I did look into the keycloak documentation. I was thinking we could do something like the below code.
```
if [ "$(opt/jboss/keycloak/bin/kcadm.sh get users -r sirixdb -q username=admin)" != '[ ]' ]; then echo 'User F
ound'; else opt/jboss/keycloak/bin/add-user-keycloak.sh -r sirixdb -u admin -p admin --roles create,modify,delete,view; fi;
```
In order to get the list of users we have to connect to the server prior to running the command. This can be achieved using the config credentials command which is available in the kcadm.sh . The below command can be used for connecting to the server.

`opt/jboss/keycloak/bin/kcadm.sh config credentials --server http://localhost:8080/auth --realm master --user admin --password admin`

Do let me know your views on this approach. Post that I can create a PR for the changes. If this is the right approach, can you also help with the parameters used for config credential command.

JohannesLichtenberger:
The script is copied into a startup folder of keycloak, so we might be able to simply copy your code: 

```
volumes:
      - ./bundles/sirix-rest-api/src/test/resources/realm-export.json:/opt/keycloak/realm-export.json
      - ./bundles/sirix-rest-api/src/test/resources/create-sirix-users.sh:/opt/jboss/startup-scripts/create-sirix-users.sh
```

MariannaDima:
Hello, can I give it a try as this is still an open issue?

JohannesLichtenberger:
Still open, also updating the Keycloak version.

PR #546, opened 2022-11-05
Fix For issue #541

Aditya31398:
Issue #541
Added Check admin user exists before creating new admin user


JohannesLichtenberger:
I think the user is not created anymore (downloaded and checked the log): 

```
2022-11-05T14:14:08.6137375Z     14:14:08.594 [vert.x-eventloop-thread-1] TRACE io.vertx.ext.web.RoutingContext - Calling the  handler
2022-11-05T14:14:08.6137830Z 
2022-11-05T14:14:08.6138140Z Integration test > Testing a simple query to get the node-key of a node STANDARD_ERROR
2022-11-05T14:14:08.6138852Z     Exception in thread "vert.x-eventloop-thread-1 @coroutine#11" io.vertx.core.impl.NoStackTraceThrowable: invalid_grant: Invalid user credentials
```

JohannesLichtenberger:
Would be awesome if you can fix this -- thanks so far for your effort :)

JohannesLichtenberger:
@Aditya31398 still the same error, sadly

Aditya31398:
When I try locally the kcadm.sh command gives out error connection refused. This is the problem. I'll try to fix it

Aminmalek:
I had this issue just one time I think Its related to how docker save the credentials and why It just happen one time, I think you need to check the ENVs another time it seems both docker and the keycloak script attempts to save the user and they have conflicts with each other.

Issue #549, opened 2022-11-11
Make sirix server wait untill keycloak is completely up and running when using docker compose


redwanulsourav:
The Sirix server depends on Keycloak for authentication. It takes a while for the Keycloak server to be up and running when using `docker compose`. The server throws exception trying to access Keycloak, and keeps on restarting. The server should not be started until Keycloak is completely up and running.

This is not specifically a bug, but it could be considered as an improvement.

JohannesLichtenberger:
Regarding Keycloak, I'm also not sure if we can switch to a newer version. I think at one point they more or less completely rewrote Keycloak for faster startup times? I'm not sure, though... will have to check.

PR #550, opened 2022-11-11
Add healthcheck to verify if keycloak is up before running server.

redwanulsourav:
Added healthcheck to make sirix server wait until keycloak is completely up and running. Fixes sirixdb/sirix#549


JohannesLichtenberger:
Thanks a lot :+1:

PR #551, opened 2022-11-12
Update documentation of setting up keycloak and server using docker compose

redwanulsourav:
Updated the instructions on setting up keycloak and server using docker compose

- Seperated the instructions for setting up keycloak using `docker compose` and docker hub image sperately.
- Updated the instructions for `keycloak.url` field. It should be set to `http://host.docker.internal:8080/auth/realms/sirixdb`. Using localhost / 127.0.0.1 cannot reach the keycloak container in local environment. 
- Structured and moved some instructions around to make things a little bit more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,6 @@
 
 >"Remember that you're lucky, even if you don't think you are, because there's always something that you can be thankful for." - Esther Grace Earl (http://tswgo.org)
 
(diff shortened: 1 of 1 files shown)


redwanulsourav:
Made some unwanted changes, will change the file accurately more and then create a pull request again.

PR #557, opened 2022-11-15
Change io_uring configuration

ikorennoy:
I suggest using this jasyncfio configuration for now.

In this configuration, using io_uring the testChicago on my machine is consistently slightly (about 20 sec) faster than the other I/O backend.

To further optimize, I would go in the direction of reducing allocation per operation, namely to implement reading via io_uring ring buffer feature.

Also, updated jasyncfio to the latest version.

### DIFF
--- a/bundles/sirix-core/src/main/java/org/sirix/io/iouring/IOUringStorage.java
+++ b/bundles/sirix-core/src/main/java/org/sirix/io/iouring/IOUringStorage.java
@@ -48,7 +48,9 @@ public final class IOUringStorage implements IOStorage {
   private final AsyncCache<Integer, RevisionFileData> cache;
 
   private static final EventExecutor eventExecutor =
-      EventExecutor.builder().entries(1024).ioRingSetupSqPoll(1_000).build();
+      EventExecutor.builder()
+              .entries(1024)
+              .build();
 
   private AsyncFile dataFile;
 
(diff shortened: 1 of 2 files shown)


ikorennoy:
Also, I wanted to add that during the test I observed that the write speed to the disk was around 10-12MB/sec. This is quite a small value for an average SSD disk. Therefore, as a benchmark for further optimizations of the I/O layer, I suggest to use and increase this characteristic.

JohannesLichtenberger:
Wow, I checked with 1024 entries before, but I think after switching to a final static instance we could even use 4096 entries, the maximum? Not sure, though, but IIRC it was a bit slower (but maybe I also switched to 4096 entries).

Regarding the write speed to disk, do you have any suggestions, which might impact this speed? The whole idea with append only / scattered reads are based on fast random read device assumptions with for instance slow in-place erase mechanisms, such that an append is more natural and better from a performance perspective.

Maybe we should simply also increase the number of nodes threshold param after which an auto-commit is issued in the test, but at some point the GC pauses are dramatically increased, that's why I think we should store the trx intent log in an ofc-heap map. Currently, I invoke `AsyncFile::write` every 65000 bytes during a trx commit and otherwise accumulate the serialized page fragments in a buffer as I read that it's not the best to write small chunks of potentially only a few hundred bytes to a few kb.

`UnorderedKeyValuePage`s are serialized in parallel, but the overhead of synchronization and CPU context switches is non-negligable, but still faster than without.

In essence the commit issues the serialization of the `UnorderedKeyValuePage`s in parallel to page-private buffers and then traverses the log in a postorder traversal, serializes and compresses other page types to a buffer to which the pages are appended and writes this buffer to disk/ to the AsyncFile (without an explicit sync) every 65000 bytes. After writing the UberPage/Header at last a sync is issued.

PR #564, opened 2022-12-14
doc: add javadoc for parseDivisionValues

ConnorCable:
Wrote a simple javadoc for a function as an example of my writing style. Feel free to give me tips on what is important in a javadoc


JohannesLichtenberger:
Thanks, can you put the text above the param and return value? Otherwise, thanks. Maybe you can drop some of the implementation details, such that we can simply change internals without invalidating the documentation (otherwise they might inadvertantly at some point diverge). Furthermore I'd rather concentrate on the public methods. However, I guess you could add comments in the code itself to get a better understanding :-)

ConnorCable:
Done, the comment is now agnostic of implementation. I'll focus on public methods!

JohannesLichtenberger:
Something I forgot: Can you also remove these hyphens between param names / return values? Thanks for the PR :+1:

PR #565, opened 2022-12-15
doc: remove hyphens

ConnorCable:
removed hyphens from parseDivisionValues javadoc


ConnorCable:
I'm not sure why this failed build

Issue #523, opened 2022-08-24
Refactor `sirix-rest-api` bundle and extract more common logic between XML and JSON stuff


JohannesLichtenberger:

redwanulsourav:
Hi, I would like to work on this issue.

JohannesLichtenberger:
Give it a try :-) the code base is rather small, but let me know if you need assistance.

JohannesLichtenberger:
@redwanulsourav did you work on the issue already?

redwanulsourav:
@JohannesLichtenberger  I am still working on it, I ran into some issues while setting up the server with docker, I think I have an idea to figure it out. Can I have some more time please?

JohannesLichtenberger:
for sure, just went through some tickets/issues to ping the people if they made some progress :+1: 

BTW: You can improve the documentation if you figured it out :-) currently, I'm mostly working on the core and didn't set up keycloak for some time now. That is usually I'm starting it via the standalone shell script ;)

redwanulsourav:
@JohannesLichtenberger I really appreciate the opportunity and the support. I was able to fix the issues I faced while setting up the keycloak and sirix server using docker. I think the setting up the sirixdb part locally could be updated. Specially the part where keycloak configuration process is described.  I also made some changes to `docker-compose.yml`. 

I would love to contribute in the documentation and improve it. :)

Currently, I was able to run the sirix server and login using `admin` account using `pysirix` client.

I still have to work on this issue itself. I might still need some more time to work on it. As I was able to run the server, I am planning to go through the code base and familiarize myself more. I really appreciate your help and support.

JohannesLichtenberger:
Yes, I have to admit, that I simply downloaded Keycloak, then executed the startup shell script and used IntelliJ for starting/debugging the server...

JohannesLichtenberger:
@redwanulsourav guess you won‘t have time to work on the issue?

redwanulsourav:
@JohannesLichtenberger It was on my radar for a few weeks now. Can you please allow me 3 more weeks? I understand I have taken a long time on this, but I should be able to work on this in the next few weeks. I am sorry for taking long time.

JohannesLichtenberger:
@redwanulsourav do you keep on working on this? Would be great

redwanulsourav:
@JohannesLichtenberger Yes, I will work on this and get the PR finished as soon as possible.

JohannesLichtenberger:
[...]

PR #586, opened 2023-03-10
Refactoring of JSONCreate and XMLCreate to extract common logic

redwanulsourav:
For #523 
Key changes:
     - Created `CreateHandler.kt` which extracts common logic between `XMLCreate()` and `JSONCreate()`
     - Currently common logic handled: Check if database already exists and if not then create it
     - Updated `SirixVerticle.kt` to use `CreateHandler` instead of `XMLCreate` and `JSONCreate`
     - Created `XMLCreate2.kt` `JSONCreate2.kt` that has mostly code from `XMLCreate.kt` and `JSONCreate.kt` so that it is usable from `CreateHandler.kt`
      
TODO:
     - Extract more common logic between `XMLCreate2` and `JSONCreate2`
     - Handle `multipart/form-data` with `XMLCreate2` and `JSONCreate2`
     - Handle and test resource creation using `CreateHandler`
     - Eventually remove `XMLCreate` and `JSONCreate`

### DIFF
--- a/bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
+++ b/bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
@@ -213,17 +213,17 @@ class SirixVerticle : CoroutineVerticle() {
             GetHandler(location, keycloak, authz).handle(it)
         }
 
-        put("/:database").consumes("application/xml").coroutineHandler {
+        put("/:database").coroutineHandler {
             Auth(keycloak, authz, AuthRole.CREATE).handle(it)
             it.next()
         }.handler(BodyHandler.create()).coroutineHandler {
-            XmlCreate(location, false).handle(it)
+            CreateHandler(location, false).handle(it)
         }
-        put("/:database").consumes("application/json").coroutineHandler {
+        put("/:database").coroutineHandler {
             Auth(keycloak, authz, AuthRole.CREATE).handle(it)
             it.next()
         }.coroutineHandler {
-            JsonCreate(location, true).handle(it)
+            CreateHandler(location, true).handle(it)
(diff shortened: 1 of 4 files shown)


JohannesLichtenberger:
I know I did something like that with few other handlers, but I think it would be worth considering if instead we could either encapsulate the common logic in abstract classes.

JohannesLichtenberger:
Thanks for working on this @redwanulsourav :+1:

redwanulsourav:
> I know I did something like that with few other handlers, but I think it would be worth considering if instead we could either encapsulate the common logic in abstract classes.

I agree, that would make it more modular. I was following the existing codebase and I thought that was how it was supposed to be done. I can implement this with abstracted classes and inheritance as well.

How it would be if I made CreateHandler a super abstract class, from which `XMLCreate` and `JSONCreate` would inherit? Later, all the common logic between handlers (`CreateHandler`, `DeleteHandler`) can be pushed upwords by creating an abstract class from which all the `Handlers` (`CreateHandler`, `DeleteHandler`) would inherit. If the `Handlers` don't have a lot in common, the super class can be turned into an interface as well. Does this plan sound ok?

JohannesLichtenberger:
Yes, sounds like a good plan (though I think maybe it's useful to have different abstract classes from which the XML and JSON implementations inherit for the CRUD stuff). `XmlCreateHandler extends AbstractCreateHandler`, `JsonCreateHandler extends AbstractCreateHandler`... and analogous the other types. I think it's better to stop at this point instead of to refactor to one big abstract class, but we'll see :-)

redwanulsourav:
I will do that and create a new PR. For now I am closing current PR. Thanks for the instructions.

JohannesLichtenberger:
Thanks for your work and interest in the project!

PR #590, opened 2023-03-11
Rest api refactor

redwanulsourav:
Refactored the code for database and resource creation. #523

### DIFF
--- a/bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
+++ b/bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
@@ -213,18 +213,19 @@ class SirixVerticle : CoroutineVerticle() {
             GetHandler(location, keycloak, authz).handle(it)
         }
 
-        put("/:database").consumes("application/xml").coroutineHandler {
+        put("/:database").coroutineHandler {
             Auth(keycloak, authz, AuthRole.CREATE).handle(it)
             it.next()
         }.handler(BodyHandler.create()).coroutineHandler {
-            XmlCreate(location, false).handle(it)
+
+            CreateHandler(location, false).handle(it)
         }
-        put("/:database").consumes("application/json").coroutineHandler {
-            Auth(keycloak, authz, AuthRole.CREATE).handle(it)
-            it.next()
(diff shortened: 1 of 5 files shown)


redwanulsourav:
@JohannesLichtenberger I kept the `handle()` functions in `XMLCreate2` and `JSONCreate2` commented out, because initially that's what I wrote, but later realized if I put the common `handle()` code in `CreateHandler.handle()` then I would have less code duplication. Similar stuff can be done for some other functions as well I believe. 

Also I think for Creation, I couldn't find many common stuff among XML and JSON that can be refactored into abstract class. Before proceeding further, I wanted to check if this was ok with you. Would it be okay to put some common logic in `CreateHandler` class? In another design, `CreateHandler` can be minimal if it simply called the `handle` function in `XMLCreate2` and `JSONCreate2`. I wasn't sure about which one was better design, so I kept it like this for now.

JohannesLichtenberger:
@redwanulsourav I would delete `CreateHandler` and just call `XMLCreate2` / `JSONCreate2` similar to how it's in the master branch (with the exception, that these classes extend the abstract class, which handles common stuff). Furthermore, I'd introduce an interface for the methods, then the common stuff in the abstract classes and specific stuff in `XMLCreate2` and `JSONCreate2` (obviously your changes have to overwrite the existing classes once we're ready to process with merging your PR) and the two new classes have to be deleted afterwards.

I'd proceed like this as well for the other classes (interface + abstract class + implementations).

redwanulsourav:
@JohannesLichtenberger I think I understood what you have mentioned. I just wanted to confirm it again. Would it be like this:`AbstractCreateHandler implements HandlerInterface` and `AbstractDeleteHandler implements Handler Interface`. From there, `XMLCreate extends AbstractCreateHandler` and `JSONCreate extends AbstractCreateHandler`. And, from there just create in `SirixVerticle` when calling, I would call `XMLCreate.handle()` or `XMLDelete.handle()` and so on.

JohannesLichtenberger:
Yes :-)

redwanulsourav:
I think `(interface + abstract + implementations)` approach will be a clean one. It is pretty close to what I have implemented so far. I am closing this PR and will create a new PR soon. Thanks for the clear instructions.

PR #592, opened 2023-03-14
Refactor REST API for Create

redwanulsourav:
- Added HandlerInterface.kt
- Added AbstractCreateHandler.kt
- Modified JsonCreate.kt to inherit from AbstractCreateHandler
- Modified XmlCreate.kt to inherit from AbstractCreateHandler


JohannesLichtenberger:
Should we close this PR?

Issue #517, opened 2022-08-24
Update Vert.x to newest version


JohannesLichtenberger:
Some stuff has been deprecated, so it might not be that trivial, but I guess it should also be a good first issue.

tsufiyan:
Can I work on this? if yes please assign

JohannesLichtenberger:
Sure, thanks

tsufiyan:
Hi @JohannesLichtenberger
Should this issue be closed?

PR #604, opened 2023-04-04
Reordered vertx handlers

tsufiyan:
Reordered vertx handlers as suggested by the updated version of vertx 4.4.0 

[Vert.x upgrade #517]


tsufiyan:
Hi @JohannesLichtenberger, 
please proceed

bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
```diff
@@ -237,7 +240,7 @@ class SirixVerticle : CoroutineVerticle() {
         put("/:database").consumes("application/xml").coroutineHandler {
             Auth(keycloak, authz, AuthRole.CREATE).handle(it)
             it.next()
-        }.handler(BodyHandler.create()).coroutineHandler {
```
JohannesLichtenberger, line 240: You've added the _BodyHandler_ to all POST requests. However, this one is a PUT request. I wonder if we miss something!? Not sure, though, as the tests pass, and the tests, as far as I can remember, issue a PUT when a resource is created.

JohannesLichtenberger, line 240: You've added the _BodyHandler_ to all POST requests. However, this one is a PUT request. I wonder if we miss something!? Not sure, though, as the tests pass, and the tests, as far as I can remember, issue a PUT when a resource is created.

tsufiyan:
Adding BodyHandler in PUT methods failing the test cases. 
Not failing, it gets caught in a deadlock condition. After the message `Lock acquired`

bundles/sirix-rest-api/src/main/kotlin/org/sirix/rest/SirixVerticle.kt
tsufiyan: Using `BodyHandler` globally for PUT method making PUT requests caught in a deadlock situation.
image

Happily, individual requests are working, so let me use `BodyHandler.create()` with each request instead of the global one. This way the design will be similar across methods and each request will have more control over itself.

JohannesLichtenberger: Sorry, can't follow... meaning you'll change the PR?

tsufiyan: Yes, we can close this one.

PR #636, opened 2023-07-07
Fix/docker network

Aminmalek:
fix the localhost in sirix conf
containers within the same network cannot directly access each other using "localhost".


Aminmalek:
I think we need 2 conf files one for docker and other one for local

JohannesLichtenberger:
Thank you!

PR #676, opened 2023-10-27
fix: fixed sonar issues

smirti-barnwal:


smirti-barnwal:
Issue:[440](https://github.com/sirixdb/sirix/issues/440)

Issue #707, opened 2024-03-16
Fix docker compose up in sirix-rest-api


JohannesLichtenberger:
Fix docker compose up regarding the Github Action (currently the tests fail)...

traitsisgiorgos:
hi, can you give same info for the problem?

JohannesLichtenberger:
The Github Action currently fails, because something apparently fails during the Docker container start, but I'm not sure how to log the error and locally it starts fine: https://github.com/sirixdb/sirix/actions/runs/8305202119

JohannesLichtenberger:
```
johannes@luna:~/IdeaProjects/sirix/bundles/sirix-rest-api/src/test/resources$ sudo docker compose -f docker-compose.yml --ansi never -p sirix-test up -d
 Container sirix-test-keycloak-1  Starting
 Container sirix-test-keycloak-1  Started
```

The stack trace in https://productionresultssa1.blob.core.windows.net/actions-results/ff8681a6-8947-40d6-9379-fac9fc03909c/workflow-job-run-5264e576-3c6f-51f6-f055-fab409685f20/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-03-17T11%3A41%3A57Z&sig=6cHBO4TinZxf6e6lnLP0Nwm4ot%2FCeY2i1PgS2yIgjx4%3D&sp=r&spr=https&sr=b&st=2024-03-17T11%3A31%3A52Z&sv=2021-12-02:

```
2024-03-17T04:09:05.0425707Z 2024-03-17T04:09:04.433+0000 [INFO] [org.gradle.process.internal.DefaultExecHandle] Starting process 'command 'docker''. Working directory: /home/runner/work/sirix/sirix/bundles/sirix-rest-api Command: docker compose -f /home/runner/work/sirix/sirix/bundles/sirix-rest-api/src/test/resources/docker-compose.yml --ansi never -p sirix-test up -d
2024-03-17T04:09:05.0429169Z 2024-03-17T04:09:04.433+0000 [DEBUG] [org.gradle.process.internal.DefaultExecHandle] Changing state to: STARTING
2024-03-17T04:09:05.0497857Z 2024-03-17T04:09:04.435+0000 [DEBUG] [org.gradle.process.internal.DefaultExecHandle] Waiting until process started: command 'docker'.
[...]

PR #710, opened 2024-04-03
gradle-version_updated

dhrax21:
1 . Updated the gradle version to 8.7 in build.gradle
2.  Updated the java version to 21 because it's showing error with version 22  the error -> "Unsupported Java. Your build is currently configured to use Java 22 and Gradle 8.7"  
3.  Updated the property in sonar -> "property ("sonar.gradle.skipCompile", true)"


JohannesLichtenberger:
Thanks, before merging fixing this is of utmost importance: https://github.com/sirixdb/sirix/issues/707    then we can be sure, that all tests run fine :-)

JohannesLichtenberger:
You'd have to update your branch from main.

Issue #711, opened 2024-04-08
Update Keycloak to newest version


JohannesLichtenberger:
The container doesn't start using the newest image.

JohannesLichtenberger:
This has prio 1 currently, as the tests fail again.

Locally 

```
johannes@luna:~/IdeaProjects/sirix/bundles/sirix-rest-api/src/test/resources$ sudo docker compose up
[sudo] password for johannes:            
WARN[0000] /home/johannes/IdeaProjects/sirix/bundles/sirix-rest-api/src/test/resources/docker-compose.yml: `version` is obsolete 
[+] Running 0/1
 ⠦ keycloak Pulling                                                       18.7s 
[DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of quay.io/keycloak/keycloak:7.0.1 to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/
```

Thus, I think it's finally time to update the version...

anmol797:
docker-compose up  is the cmd right ? to test

JohannesLichtenberger:
Right, but the startup scripts are not loaded anymore plus you can look into the docker log, I guess

anmol797:
Hi 
anmol797:main the container is up and running now 
https://github.com/sirixdb/sirix/pull/729 here is the updated changes 
please check once

PR #717, opened 2024-04-20
Changed and added commands in docker files

FayKounara:


JohannesLichtenberger:
Oh, what's this about? I thought it's Keycloak updated somehow

FayKounara:
Hello! So these changes help the container keycloak to not start and exit immediately and when i altered these i was also able to build successfully the build.gradle from the folder sirix-rest-api.

JohannesLichtenberger:
The thing is that the docker compose file in the root dir should build a new image from the sources (afterwards it's pushed to docker hub for use by the rest api bundle and for everyone).

FayKounara:
Do you mean the sirixdb/sirix image?

docker-compose.yml
```diff
@@ -2,30 +2,34 @@ version: '3'
...
     networks:
       - auth-network
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8080"]
+      test: "curl --fail  http://localhost:8080 || exit 1"
       interval: 30s
       timeout: 5s
       retries: 10
   server:
-    build: .
```
JohannesLichtenberger, line 28: This should be in, as it's automatically built after successful tests and pushed t docker hub :-)

FayKounara:
Okay so i added this command and executed it again.From the reports that are produced i see this. Package | Tests | Failures | Ignored | Duration | Success rate
-- | -- | -- | -- | -- | --
io.sirix.rest | 66 | 0 | 0 | 19.163s | 100%
So can you please explain what is the problem with the sirix-rest-api because i see all the containers are running smoothly. Maybe i dont understand something. Thank you and sorry for the trouble.

JohannesLichtenberger:
The docker compose file in sirix-rest-api uses the most recent version from docker hub, which is built before via the docker compose file in the root directory of the project. I'm still not really sure what's the problem you're facing. So, using the docker compose script in sirix-rest-api starts keycloak and stops it immediately?

FayKounara:
Sorry for the confusion!  I saw on the issue #711 this,The container doesn't start using the newest image,and also on the issue #707 that the tests fail. I think that these changes might help with these issues.

JohannesLichtenberger:
Oh, #707 was fixed (the keycloak image only exists for quay.io). Idk what's wrong when updating Keycloak to the newest version, though. In any case, if your changes help, we could add the local build again, I guess, regarding the root docker compose file!? Otherwise it wouldn't be updated anymore (and pushed to docker hub).

FayKounara:
If my changes prove effective, I will add the local build option in the root Docker Compose file again.Thank you and sorry for the confusion!

JohannesLichtenberger:
So, you'll update the keycloak version and check?

FayKounara:
Yes i will look into it.

JohannesLichtenberger:
You'd have to update the Keycloak version as well (also add `build .` again, as it's mandatory to do a local image build)

JohannesLichtenberger:
Can welcome this?

FayKounara:
Hello, yes!

Issue #537, opened 2022-10-13
Support for realtime web 


JohannesLichtenberger:
Meaning in a first step we should stream the data changes for a resource to interested clients (we could just stream the JSON we generate, which stores a change track).

In a second step we can support queries, which only query one resource. Thus, naively we'd probably have to execute the query/queries asynchronous once a trx has been flushed to disk and check that only result nodes in the curre transaction intent log are streamed to interesting clients. Maybe we can store the compiled query at least up until indexes are added/dropped.

Better ideas are of course welcome. A simpler much more efficient approach would be to check for updates on certain paths in the resource.

JohannesLichtenberger:
How does RethinkDB implement this?

mosheduminer:
I wonder if looking at https://github.com/pubkey/event-reduce would be helpful.

JohannesLichtenberger:
Thanks Moshe, do yohh understand the basic algorithm they are using? Maybe it's too late currently, but I don't get how they determine if a new record from a change event is a query result or not.

mosheduminer:
I don't know, I haven't really looked at the implementation. So I only know what it says in the readme.

ElenaSkep:
Hello! I am interested in working on this issue. Would it be ok? I was thinking of using webSockets to stream the json. Also could you specify in which path I should focus first?

JohannesLichtenberger:
You can start by streaming updates to subscribed clients: https://github.com/sirixdb/sirix/blob/982a346f6dbceeade7cf581199c47c910804f14d/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractNodeTrxImpl.java#L297

Maybe we can add a post commit hook to listen for changes.

JohannesLichtenberger:
So, I think it would be great if clients can subscribe to a database/resource update stream, thus that they receive what has been changed as a JSON change stream (we already write these changes to JSON files on disk).

So we need to have a simple pub/sub mechanism, checking for read-access rights and to write the changes to the websocket.

ElenaSkep:
Hello again! @FayKounara and I have made some progress. We have created a pub/sub mechanism using Apache Kafka but it would be helpful if you could specify who is considered an interested client. Also we are using WebSockets and Nginx if its ok.

JohannesLichtenberger:
[...]

PR #726, opened 2024-06-12
Issue #537

ElenaSkep:
A websocket client detects the changes made in a database/resource and sends them to a websocket server.On the console where you run sirix you can see the change that was made. What you can do is create a Websocket server (we created one with node.js) which receives the changes from the db and can broadcast them to all connected clients (we implemented those also with node.js).
It should be noted that we do not provide the code for the server and the clients that we used to run it.

### DIFF
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/json/JsonNodeTrxImpl.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/json/JsonNodeTrxImpl.java
@@ -19,2698 +19,2742 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.sirix.access.trx.node.json;
-
-import com.google.gson.stream.JsonReader;
-import com.google.gson.stream.JsonToken;
-import io.brackit.query.atomic.QNm;
-import io.brackit.query.jdm.Item;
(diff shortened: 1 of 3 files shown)


ElenaSkep:
I would like to state that in JsonNodeTrxImpl.java our changes are 28 lines (2732-2759) and the rest of the code is the same.

JohannesLichtenberger:
A direct websocket dependency sadly doesn't make much sense, as `sirix-core` along with other bundles should be usable without a server.

Instead, we already have the `sirix-rest-api` where a new route should be added along with authorization checking for read permission and along with a simple pub/sub mechanism. We have to figure out if we can add a method, which is not leaked into the API to get the updates instead of hardcoded in the method, which writes the changes to disk.

So, instead of the NodeJS based backend you used we already have a server based on Vert.x and Kotlin in `sirix-rest-api`. Can you implement this over there as a separate route as mentioned before (instead of the websocket client we'd be on the server already)?

ElenaSkep:
Thank you for taking the time to look at our idea. With the addition of websockets the sirix-core will function as before and the only difference is that the message will be able to be broadcasted .The websockets don’t interfere with the way sirix-core functions. Regarding to the sirix-rest-api it relies heavily on keycloak which still causes problems with running it.

JohannesLichtenberger:
Well, it seems as of yesterday the tests fail due to a docker image issue again. I guess it's finally time, that we have to update Keycloak and to get it up-and-running again.

PR #729, opened 2024-06-25
#711: Update Keycloak to newest version

anmol797:


JohannesLichtenberger:
"Waiting for keycloak-1 to become healthy (it's starting)"

Hmm

anmol797:
> "Waiting for keycloak-1 to become healthy (it's starting)"
> 
> 
> 
> Hmm

the java CI gradle build is also getting failed , any idea ?

anmol797:
> "Waiting for keycloak-1 to become healthy (it's starting)"
> 
> 
> 
> Hmm

where did you see that log ?

JohannesLichtenberger:
You have to check the docker container log locally, I guess and I still think the realm needs to be imported, as otherwise the admin user and the roles are not available and there's no out of the box way to add startup scripts anymore, I guess

JohannesLichtenberger:
BTW I think you should be able to check if the tests for `sirix-rest-api` run locally.

anmol797:
> BTW I think you should be able to check if the tests for `sirix-rest-api` run locally.

the container is not coming up , trying to use that jar method now 
then will run the IT

JohannesLichtenberger:
I guess you've had no success so far? Maybe we have to create our own Keycloak image to still be able to run startup scripts...

anmol797:
> I guess you've had no success so far? Maybe we have to create our own Keycloak image to still be able to run startup scripts...

how to make it ?

JohannesLichtenberger:
Going to close this for now

Issue #465, opened 2022-01-12
Git hook to autoformat source code


JohannesLichtenberger:
We need a pre commit hook or something like that to auto-format code according to our IntelliJ Java Formatter or the Kotlin standard code formatting rules.

omegaspard:
Can I work on this ? To be honest I'm reading the project since a few day and I really would like to participate. I'm new to open source contribution, I don't even know if I have to ask that way.

JohannesLichtenberger:
Sure, I'll have to upload the formatter later on.

gvdutra:
Welcome omegaspard :)

abhishekscte0327:
Hello, is this issue still open? I would like to work on it.

abhishekscte0327:
#### Issue with Spotless Configuration

Hi @Johannes Lichtenberger,

I'm currently working on a contribution to the project and encountered an issue with the `spotless` configuration. The pre-commit hook for auto code formatting is failing with the following error:
```groovy
Task :sirix-core
FAILED

Step 'com.diffplug.spotless.generic.LicenseHeaderStep' found problem in 'src/test/java/io/sirix/index/art/AbstractNavigableMapShortTest.java':
Unable to find delimiter regex ^(package|import|public|class|module)
java.lang.IllegalArgumentException: Unable to find delimiter regex ^(package|import|public|class|module)
```

## Root Cause

This issue seems to be caused by the following line in the `spotless` configuration:

```groovy
licenseHeaderFile "${rootDir}/LICENSE"
```

**It appears that the problem arises because the project contains files that are fully commented. The spotless tool is unable to find the expected delimiter regex in these files, leading to the formatting error.**

After commenting out the licenseHeaderFile line in the spotless configuration, the pre-commit hook for auto code formatting runs fine.

However, I would prefer to find a more permanent solution that does not require removing the license header. I have reached out to you on the Discord channel to discuss this issue and explore potential solutions.


Looking forward to your guidance

JohannesLichtenberger:
Implemented. There is a `pre-commit` hook at the repo root that runs formatting, and the build applies the Spotless plugin (`com.diffplug.spotless`) in `build.gradle` (`subprojects { apply plugin: "com.diffplug.spotless" ... eclipse().configFile("${rootDir}/eclipse/sirix-formatter.xml") }`). Closing as done — reopen if you'd like the hook wired up automatically on clone (e.g. via a Gradle `installGitHooks` task).

---
[...]

PR #734, opened 2024-08-22
enhancement: Automate Git hook setup for code formatting with Spotless

abhishekscte0327:
- Added githooks configuration to automate code formatting using Spotless.
- Ensured that the hooks are set up automatically when running `./gradlew` commands.
- This enhancement will help maintain consistent code formatting across the project.

This PR addresses issue <https://github.com/sirixdb/sirix/issues/465>.


JohannesLichtenberger:
@abhishekscte0327 we should use this formatter: https://github.com/sirixdb/sirix/blob/main/SirixDB.xml

abhishekscte0327:
Hii @JohannesLichtenberger, 
I've run into an issue with the provided SirixDB.xml file. It’s specific to IntelliJ IDEA and isn’t compatible with tools like Spotless or Checkstyle, which are cross-IDE friendly. Translating the provided xmlinto a format like Eclipse or Checkstyle is possible but not all rules can be perfectly mapped.

Would you prefer we go with an IntelliJ-specific solution, or should we explore a cross-IDE approach? I'm open to suggestions if you have any other ideas.

JohannesLichtenberger:
IIRC, the file was exported from Eclipse originally; maybe you can trace it back :-) That said, maybe we can otherwise go for an IntelliJ-specific solution, as reformatting almost everything, IMHO may not be worth it, as it also destroys history.

abhishekscte0327:
I have observed that using IntelliJ-specific formatting also results in reformatting a large portion of the codebase. However, i successfully traced the XML file to its original Eclipse format.

We can proceed with a cross-IDE solution using Spotless with the Eclipse formatter.  Please note that this approach will still lead to significant reformatting of the codebase.

**As an alternative, i propose using a pre-commit hook that formats only the modified files. This solution minimizes disruption by incrementally applying formatting changes, preserving the commit history, and avoiding a single massive reformatting commit. Here’s a brief overview of the approach:**

**Pre-Commit Hook**: Formats only the files that are staged for commit.
**Gradual Formatting**: Ensures formatting is applied over time, reducing the impact on the codebase.
Please let me know if you’re comfortable with this approach or if you have any other suggestions.

JohannesLichtenberger:
Sounds great, I'm not sure if it had "proper" formatting of streams back then, as I mostly prefer, when each function call is aligned on the next line, but maybe only if they wouldn't fit on one line.

Otherwise the approach sounds good :-)

abhishekscte0327:
Thanks for your feedback!

Regarding the formatting of streams where each function call is aligned on the next line, we can customize the formatting rules to fit this preference. We can ensure that function calls are aligned properly, either fitting them on one line or aligning them on the next line when they don't fit.

I'll proceed with setting up the pre-commit hook and the cross-IDE formatting solution using Spotless with the Eclipse formatter. We can tweak the settings further if needed to match the preferred formatting style.

Please let me know if you have any specific formatting configurations in mind, and I'll incorporate them.

JohannesLichtenberger:
This can be closed, right?

abhishekscte0327:
yeah, i am closing it.

PR #764, opened 2025-05-23
CoroutineDescendantAxis.kt revamp.

xDido:


JohannesLichtenberger:
Hey, thanks a lot, but I think we need to first check if the right sibling is on the same page (meaning it'll be cached already). Only if it's not we can use another trx to fetch it in the meantime.

In general I'm not sure if synchronization overhead might be way too much, I don't really know, but I guess it might be.

JohannesLichtenberger:
Oh, saw that you did this, but you can use `PageReadOnlyTrx::pageKey` to get the page number of a node.

JohannesLichtenberger:
BTW: the integration tests fail

JohannesLichtenberger:
You also most likely get the descendantCount of nodes directly. Furthermore, you can navigate with `moveToFirstChild()`, `moveToRightSibling()`. Waiting 1s for a parallel result is also way too long and if there's nothing in the queue there's nothing. Also, I think you wouldn't use a BlockingQueue in Kotlin. You typically use a `Channel` instead.

I don't know why your logger seems to be null, but maybe you're missing a configuration?

Over here the precondition fails? https://github.com/sirixdb/sirix/blob/5b8dd63137e55ffe20358f93512edd81ef67f973/bundles/sirix-core/src/main/java/io/sirix/utils/LogWrapper.java#L45

xDido:
I have added a logging message in the LogWrapper constructor as well as throughout the entire class. I ran the tests, and nothing in the stack trace indicates any issues originating from the log wrapper. It simply states that it's null on the first call of log.debug at line 105 in the reset function.  
image

PR #960, opened 2026-04-28
perf: prefetch-pipelined temporal axes (AllTime / Past / Future)

JohannesLichtenberger:
## Summary

The multi-revision temporal axes (`AllTimeAxis`, `PastAxis`, `FutureAxis`) walk revisions strictly sequentially. Each `computeNext` call does a fresh `beginNodeReadOnlyTrx` — full UberPage navigation, RevisionRootPage load, indirect-page traversal — before yielding to the consumer. Cores spent opening revision N+1 while the consumer processes revision N are otherwise idle.

This PR adds **prefetch-pipelined sibling axes** that overlap those opens with consumer work using a fixed-depth virtual-thread queue.

## What changed

- **`RevisionPrefetcher`** — shared lookahead engine. Maintains an `ArrayDeque<CompletableFuture<RtxResult>>` sized to `depth=4` by default. Each task starts a **virtual thread directly** (no executor lifecycle, no shutdown coordination). `poll()` blocks on the head future via `.join()` and tops the pipeline up after each pop. `close()` attaches a continuation that closes any `rtx` whose open completes after the axis is abandoned — leak-free even if the consumer drops the iterator mid-stream.
- **`PrefetchedAllTimeAxis` / `PrefetchedPastAxis` / `PrefetchedFutureAxis`** — drop-in API-compatible siblings of the sequential axes. **Iteration semantics match byte-for-byte** (same yielded sequence, same termination rules). Each supplies the prefetcher with a direction-specific revision iterator.
- **Bonus correctness fix** in `PrefetchedAllTimeAxis`: closes `rtx`s whose `moveTo` fails before the first yielded revision — the original `AllTimeAxis` leaked them on the prefix scan.

## What's not parallelized

`FirstAxis`, `LastAxis`, `NextAxis`, `PreviousAxis` open exactly one `rtx`. Nothing to parallelize.

## HFT properties

- Bounded memory: at most `depth + 1` rtx in flight at any moment.
- No long-lived executor — virtual threads are spawned per task. No per-axis pool, no shutdown.
- Backpressure-free reads: `poll()` blocks on the head future; consumer processing time gates the prefetch rate naturally.
- Leak-free shutdown: pending futures close their rtx via `thenAccept`, even after the axis exits.

## Test plan

- [x] `PrefetchedAllTimeAxisTest`, `PrefetchedPastAxisTest`, `PrefetchedFutureAxisTest` — mirror the existing sequential-axis test suites and verify identical yielded sequences across the look-ahead pipeline (incl. deleted-node termination).
- [x] Existing `AllTimeAxisTest`, `PastAxisTest`, `FutureAxisTest` stay green (sequential axes untouched).
- [ ] CI on PR

### DIFF
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
@@ -861,6 +861,14 @@ public abstract class AbstractResourceSession<R extends NodeReadOnlyTrx & NodeCu
   }
 
   /**
+   * Number of currently-open read-only or read-write node transactions on this session.
+   * Useful for diagnostics and for asserting the absence of resource leaks in tests.
+   */
+  public int activeTrxCount() {
+    return nodeTrxMap.size();
+  }
+
+  /**
    * Begin a read-only transaction at the revision that was valid at the given point in time.
    * 
    * <p>
--- a/bundles/sirix-core/src/main/java/io/sirix/api/ResourceSession.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/api/ResourceSession.java
@@ -95,6 +95,14 @@ public interface ResourceSession<R extends NodeReadOnlyTrx & NodeCursor, W exten
   Optional<W> getNodeTrx();
 
   /**
+   * Number of currently-open read-only or read-write node transactions on this session.
+   * Useful for diagnostics and for asserting the absence of resource leaks in tests.
+   *
+   * @return the number of open node transactions, including the single writer if any
+   */
+  int activeTrxCount();
+
+  /**
    * Begin a new {@link StorageEngineReader}.
    *
    * @return new {@link StorageEngineReader} instance
--- a/bundles/sirix-core/src/main/java/io/sirix/axis/AbstractTemporalAxis.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/axis/AbstractTemporalAxis.java
@@ -9,11 +9,24 @@ import io.sirix.utils.AbstractComputingIterator;
 /**
  * TemporalAxis abstract class.
  *
- * @author Johannes Lichtenberger
+ * <p>Implements {@link AutoCloseable} so consumers (typically a wrapping
+ * {@code TemporalSirix*Stream}) can release any resources the axis holds — most
+ * importantly, prefetched read-only transactions held by look-ahead variants. The
+ * default {@link #close()} is a no-op for axes that hold no per-iteration resources.
  *
+ * @author Johannes Lichtenberger
  */
 public abstract class AbstractTemporalAxis<R extends NodeReadOnlyTrx & NodeCursor, W extends NodeTrx & NodeCursor>
-    extends AbstractComputingIterator<R> {
+    extends AbstractComputingIterator<R> implements AutoCloseable {
 
   public abstract ResourceSession<R, W> getResourceSession();
+
+  /**
+   * Release any resources this axis holds. Safe to call multiple times. The default
+   * implementation does nothing; look-ahead axes override to cancel pending prefetches.
+   */
+  @Override
+  public void close() {
+    // no-op default
+  }
 }
--- a/bundles/sirix-core/src/main/java/io/sirix/axis/temporal/AllTimeAxis.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/axis/temporal/AllTimeAxis.java
@@ -52,8 +52,13 @@ public final class AllTimeAxis<R extends NodeReadOnlyTrx & NodeCursor, W extends
       if (rtx.moveTo(nodeKey)) {
         hasMoved = true;
         return rtx;
-      } else if (hasMoved) {
-        rtx.close();
+      }
+      // moveTo failed: either we're still in the prefix before the node was created, or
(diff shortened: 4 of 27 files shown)


JohannesLichtenberger:
## Follow-up review (auto-run 2026-05-06)

> **Context:** This PR merged on 2026-04-29, same day as the original review. All findings below are against merge commit `2c56db5b` (the final state now in `main`). No changes were requested pre-merge, so this comment serves as a post-merge record.

---

### Resolved

- **(b) `RevisionPrefetcher.close()` — cancellation strategy.**
  `close()` calls `f.cancel(true)` on every queued future, followed by `f.whenComplete(CLOSE_RESULT_RTX)` to handle futures that already completed normally. The supplier body checks the `volatile closed` flag at **two** cooperative cancellation points (before opening the trx, and immediately after the `synchronized` block returns). The decision to use the `closed` flag rather than `Thread.currentThread().isInterrupted()` is explicitly and correctly documented: `CompletableFuture.cancel(true)` does not actually interrupt the `supplyAsync` body under Java's contract. Design is correct and well-explained. ✓

- **(c) `AutoCloseable` wiring is complete.**
  `AbstractTemporalAxis` now implements `AutoCloseable` with a documented no-op default. All three prefetch axes (`PrefetchedAllTimeAxis`, `PrefetchedPastAxis`, `PrefetchedFutureAxis`) override `close()` to call `prefetcher.close()`. Both stream wrappers — `TemporalSirixNodeStream` and `TemporalSirixJsonArrayStream` — delegate to `axis.close()` in their own `close()` methods. The chain is unbroken. ✓

- **(d) Constructor is now lazy.**
  `RevisionPrefetcher` constructor sets fields only; `fillToDepth()` is called on the first `poll()`. Confirmed by code inspection and pinned by `RevisionPrefetcherLifecycleTest.constructorIsLazy_noOpensSubmitted`, which asserts `inFlight() == 0` and `supplier call count == 0` after construction. ✓

- **(e) Javadoc no longer overclaims pipeline depth.**
  The class-level Javadoc in `RevisionPrefetcher` explicitly states: *"`beginNodeReadOnlyTrx` is `synchronized` on the resource session, so its body … executes one task at a time."* The claim is accurately narrowed: the depth-N pipeline pays off on the **`moveTo`** traversal (parallel across different revisions' page trees), not on the serialized open. No mention of "depth-4 overlap" for the open phase. ✓

- **(g) Abandonment-path tests are present and meaningful.**
  `RevisionPrefetcherLifecycleTest` covers: lazy construction, close-before-any-poll (constructor-without-iteration), close-after-first-poll (mid-iteration abandon), poll-after-close returns null, exhausted-iterator, negative-depth rejection, and null-argument rejection. `AllTimeAxisTest.prefixRtxIsClosed_noLeak` uses `activeTrxCount()` to assert no transactions leak on the prefix scan. `PrefetchedAllTimeAxisTest`, `PrefetchedPastAxisTest`, and `PrefetchedFutureAxisTest` are present and mirror the sequential-axis suites. ✓

---

### Outstanding

- **(a) `AbstractResourceSession.beginNodeReadOnlyTrx` is still `synchronized`.**
  At merge commit `2c56db5b`, the method signature remains `public synchronized R beginNodeReadOnlyTrx(final int revision)`. The `nodeTrxMap` was upgraded to `ConcurrentHashMap`, which is correct for concurrent read access, but the method body as a whole still acquires the intrinsic lock on `this`. The practical effect is that all four depth-4 supplier tasks queue at this monitor before each can proceed to `moveTo` — so the true parallelism is limited to concurrent `moveTo` traversals, not concurrent opens. The Javadoc now honestly documents this (see item (e) above), so the **claim** is accurate. But the bottleneck itself has not been removed. A `ReentrantReadWriteLock` (write-lock for `beginNodeTrx`, read-lock for `beginNodeReadOnlyTrx`) or dropping to `ConcurrentHashMap`-based ID generation without any monitor would unlock the full pipeline.

  **JMH note:** Because `beginNodeReadOnlyTrx` is still serialized, the benchmark described in the original review would confirm a real but partial speedup — `moveTo` parallelism is meaningful for deep histories with cold page caches — but it would not change the structural conclusion. Skipping the benchmark run accordingly.

- **(f) `AllTimeAxis` prefix-leak fix bundled in this PR rather than split.**
  The fix landed in its own commit (`2c56db5b27b`, *"fix: AllTimeAxis prefix-rtx leak; add ResourceSession.activeTrxCount()"*), clearly scoped and well-tested. That's the right shape for a fix commit. However, the original review asked for it to be split into a **separate PR** against `AllTimeAxis` so the correctness fix could be merged independently of the prefetch feature. It was not — it rode in the same PR. Since the PR is already merged this is moot for this branch, but worth noting if the pattern recurs: pure correctness fixes to existing classes should travel as their own PR so they can be cherry-picked to older release branches if needed.

---

### Recommendation

**No further action required on this PR (already merged).** Consider a follow-up issue or PR to narrow the `synchronized` scope on `beginNodeReadOnlyTrx` — replacing the method-level intrinsic lock with a `ReentrantReadWriteLock` (read-locked for read-only opens, write-locked for `beginNodeTrx` and `close`) would let all four prefetch tasks acquire their trx concurrently and make the depth-4 pipeline claim fully pay out. The rest of the design — lazy pipeline, cooperative cancellation, AutoCloseable wiring, honest Javadoc, and abandonment tests — is solid and production-ready.

---

PR #966, opened 2026-05-01
Feat/temporal axis prefetch

JohannesLichtenberger:


JohannesLichtenberger:
Companion Brackit PR: sirixdb/brackit#95 widens \`Duration.days\` from \`short\` to \`int\` to fix a silent 89-year truncation in \`xs:dateTime\` subtraction. Recommended merge order:

1. Merge sirixdb/brackit#95.
2. Bump the Brackit dependency in this repo (or just re-install \`1.0-SNAPSHOT\` locally for CI).
3. Merge this PR.

Sirix-side impact is zero — no consumer here calls \`Duration.getDays()\` directly, so the return-type widening is source-compatible. Bitemporal & query integration tests pass against the new Brackit jar locally.

The full invariant set for this PR — including the Brackit arithmetic Inv 1.1 – 1.5 — is documented in \`docs/formal-verification.md\` with proof sketches and per-invariant test references. Property tests at 10 000 iterations each cover \`DTD\` / \`YMD\` add-subtract round-trips and \`dateTime\` subtract over a ±359-year span (post-widening: exact; pre-widening: would have wrapped silently on most iterations).

PR #1021, opened 2026-06-09
perf: fix node-history and query latency (event-loop blocking + uncached history)

JohannesLichtenberger:
## Summary

Fixes two user-visible latency problems in the GUI/REST path: **node history is slow** and **queries feel slow** — three root causes, each verified against the code:

### 1. `perf(core)`: history was O(revisions) full-trx opens, every call
`getHistory*` opened a **full `NodeReadOnlyTrx` per revision on every call** just to read four fields (revision, timestamp, author, message) off the `RevisionRootPage` — and re-did all of it on each subsequent call.

- **Session-scoped `revision → RevisionInfo` cache**: committed revisions are immutable, so entries never need invalidation. Repeated history calls only pay I/O for revisions committed since the last call. Loop bounds still come from `getMostRecentRevisionNumber()`, so new commits always appear.
- **Cache misses read via `StorageEngineReader`** instead of `beginNodeReadOnlyTrx` — credentials + timestamp live on the `RevisionRootPage`; the node-cursor machinery was pure overhead.

### 2. `perf(rest-api)`: blocking work ran ON the Vert.x event loop
- `HistoryHandler` ran the database open + per-revision retrieval + JSON assembly inside `withContext(vertx.dispatcher())` — **on the event loop**, unlike every other handler. A slow history request stalled the loop and queued **every concurrent request** behind it (including query-editor calls) — making everything feel slow, not just history.
- `AbstractGetHandler` had the same bug one level down: query evaluation/serialization already offload via `executeBlocking`, but the database open, resource-session open and revision(-timestamp) resolution ran on the event loop.

Both now run on `Dispatchers.IO` (the codebase's established pattern); responses are written back on the Vert.x context.

## Test evidence (local)
- New test `testGetHistoryIsStableAcrossCallsAndSeesNewCommits` — guards the two semantics the cache must preserve (identical metadata across calls; **new commits visible after a cached call**; bounded calls still ordered).
- `JsonMultiRevisionTest` 15/15 (incl. existing `getHistory` assertions), `UpdateTest` 40/40, `AllTimeAxisTest` 3/3, `PrefetchedAllTimeAxisTest` 5/5, `PastAxisTest` 4/4 — 0 failures.
- `sirix-core` + `sirix-rest-api` compile clean.

### DIFF
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
@@ -107,6 +107,15 @@ public abstract class AbstractResourceSession<R extends NodeReadOnlyTrx & NodeCu
   final ConcurrentMap<Integer, StorageEngineWriter> storageEngineWriterMap;
 
   /**
+   * Cache of revision → {@link RevisionInfo} (author, timestamp, commit message). A committed
+   * revision is immutable, so an entry never has to be invalidated for the lifetime of this
+   * session; repeated history calls (the REST {@code /history} endpoint, GUIs polling the
+   * revision list) only pay I/O for revisions committed since the last call instead of
+   * re-reading every {@code RevisionRootPage} each time.
+   */
+  private final ConcurrentMap<Integer, RevisionInfo> revisionInfoCache = new ConcurrentHashMap<>();
+
+  /**
    * Lock for blocking the commit.
    */
   private final Lock commitLock;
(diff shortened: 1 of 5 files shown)


JohannesLichtenberger:
Measured the core change with a local timing harness (100 revisions, warm FS — ratios are the story; a cold small box amplifies the absolute values):

| path | time |
|---|---:|
| old: full `NodeReadOnlyTrx` per revision (cost paid on **every** call) | 9.87 ms |
| new: `StorageEngineReader` per revision | 2.61 ms (3.8×) |
| new: `getHistory()` first call (parallel + cache fill) | 3.41 ms |
| new: `getHistory()` **cached** (steady state — what a polling GUI hits) | **0.09 ms (~110×)** |

Plus the event-loop fix means a history request no longer adds its latency to every concurrent request on the same event loop.

PR #1030, opened 2026-06-12
Fix HOT projection storage: grow-overwrite corruption and insert-failed-after-split

JohannesLichtenberger:
## Problem

Persisting projection-index leaves with grown payloads (e.g. a 3-column projection rebuilt to 6 columns at the same slot ids) failed with `Projection HOT chunk insert failed after split` — and worse, several variants corrupted data silently. HOT value chunking itself is by design (versioning granularity + storage economy: only changed chunks rewrite across revisions) and is fully preserved.

## Root causes (three mechanisms under one symptom)

1. **Use-after-close via stale swizzled PageReferences (primary).** `TransactionIntentLog.put` closes the replaced `HOTLeafPage` when a CoW overwrite reuses a `logKey` (correct — frees the 64 KiB frame). But CoW'd `HOTIndirectPage`s copy child-reference arrays *including the swizzled page pointer*, so deep split cascades read recycled-frame memory through stale copies: NPEs in `computeInitialBytePos`, bogus split decisions, corrupted payloads (tripwire-confirmed `closed=true orphaned=true` reads; 4,125 corrupted leaves in a 20K-leaf rebuild). Fixed centrally: `PageReference.getPage()` treats a closed HOT leaf as a cache miss so all 36 resolution sites re-resolve via logKey/disk — the fallback the TIL design already promises.
2. **Silently dropped grown-chunk updates.** Chunk writers ignored `HOTLeafPage.updateValue` returning `false` when a size-changed chunk no longer fit — old payloads survived reads. Now routed through the standard versioned leaf split (update-or-split), with a loud failure carrying leafIndex/chunkIdx context if placement is genuinely impossible.
3. **Wrong split-insert semantics + invalid rollback.** `splitToWithInsert` bitmap-merged ALL value types — opaque PROJECTION chunk payloads got "merged" as roaring bitmaps (corruption) — and its failure rollback assumed slot state was unmodified (compaction/prefix-rebuild had invalidated that). Now index-type-dispatched (`putOrReplace` for PROJECTION; CAS/PATH/NAME merge unchanged) with a full-snapshot atomic rollback.

Plus read-side fixes found while validating end-to-end: the parallel depth-2 hydrate replaced fragments on merge (multi-sub-tree leaves came back truncated); chunks are now placed at absolute offsets with fill bitmaps and a real fragment merge; tombstones truncate to match `get()`; and a fail-fast `CHUNK_SIZE` capacity contract.

## Verification

- New `ProjectionIndexHOTStorageGrowingPayloadTest` (7 tests; 4 of 5 original methods failed pre-fix with silent corruption) and `ProjectionPersistForceRebuildTest` (end-to-end 3-col → 6-col force-rebuild persist; fails pre-fix, passes post-fix).
- Re-enabled two formerly `@Disabled` tests that shared the use-after-close root cause: `HOTFormalVerificationTest.comprehensiveManyDuplicatesLowCardinality` and `scaleRandomShuffle100K`.
- Full `:sirix-core:test`: **9,194 tests, 0 failures, 0 errors**.
- Scaled repro: 20,000 leaves × 20 KB → 40 KB force-rebuild — pre-fix 2,124 exceptions + 4,125 corrupted leaves; post-fix 0/0, all leaves byte-identical.


JohannesLichtenberger:
Superseded by https://github.com/sirixdb/sirix/pull/1032 — same fix rebased onto main (this branch was CONFLICTING, so CI could not run).

PR #1052, opened 2026-06-25
docs: refresh for the beta series (HN-readiness)

JohannesLichtenberger:
Pre-Show-HN doc cleanup (the public-facing inconsistencies an HN reader would pick at):

- **ROADMAP.md** — was "currently in the **1.0.0-alpha** series"; we ship **beta4**. → "1.0.0-beta".
- **CHANGELOG.md** — was frozen at `[1.0.0-alpha10]` (14 releases stale, reads as abandoned). Added entries for `beta1`–`beta4` and a consolidated `alpha11`–`alpha22` hardening summary, with a pointer to GitHub Releases for full per-version notes.
- **README.md** — corrected "Single JAR, **no external dependencies**" (it's a shaded fat jar that bundles brackit/Jackson/SLF4J/Caffeine/RoaringBitmap), and softened the unsourced "beats jq by up to **6.8×**" to a sourced, directional claim.

Doc-only; no code changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,48 @@
 
 All notable changes to SirixDB are documented in this file.
 
+## [1.0.0-beta4] — 2026-06-19
+
+### Changed
+
+- Bumped brackit to `1.0-alpha7`, fixing the sequence functions (`fn:subsequence`/`reverse`/`remove`/`insert-before`) over JSON arrays and objects.
+
+## [1.0.0-beta3] — 2026-06-19
+
+### Added
+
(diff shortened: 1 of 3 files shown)


hiSandog:
This cleanup makes the public-facing docs read much more current. One small follow-up: the README now says brackit runs analytical JSON queries several times faster than `jq` in its own benchmark suite, which is softer than the previous claim but still benchmark-sensitive. It may be worth linking that exact benchmark suite/result near the sentence, so HN readers can inspect workload shape without hunting through the repo.

PR #1122, opened 2026-07-19
Projection-index production hardening: stress/soak suite, REST serving, three engine fixes

JohannesLichtenberger:
## What does this PR do?

Hardens the projection index toward dropping its "experimental" label, along the two axes it hadn't been exercised on: sustained incremental-maintenance load and HTTP exposure.

**1. Stress/soak suite (`ProjectionIndexStressTest`)**
- Randomized incremental-maintenance soak: 375 attributable operations across 15 commits over 2200+ rows (in-place updates, deletes, tail/middle inserts, `{}`/null rows) with a deterministic oracle checked after every commit, leaf-level physical-row bounds, time travel over every soak revision, cold re-open discovery, and a vectorized-vs-interpreted query battery with a `servedCount` proof.
- Concurrent readers (fresh database + session per check, the shape every REST request has) racing an updating writer across revisions.
- Repeated tombstone → same-definition rebuild cycles via subtree moves out of the record set.

**2. Three engine bugs the suite surfaced, all fixed**
- *Stale in-process registry serving after invalidation:* the committed lookup fell back from the revision-scoped catalog to the in-memory `ProjectionIndexRegistry`, whose handles are gated only by `validFromRevision` — a snapshot installed before an invalidating commit kept serving every later revision in the same process. The catalog is now authoritative whenever catalogued projection definitions exist at the executor's revision; the registry fallback remains for bench/test wiring only.
- *Unscoped name sweep after re-open:* the generic aggregate's target-path resolution read `PathNode#getName()` directly, which is only populated for path nodes created in the current process. On a deserialized summary every named ancestor looked anonymous, resolution failed, and the scan silently degraded to an unscoped name sweep that over-counted same-named fields under other roots. Names now resolve through the positioned `PathSummaryReader`, with the `__array__` sentinel treated as anonymous (empty names map to anonymous in both branches, so fresh and re-opened summaries walk identically).
- *Moved subtree kept its old `pathNodeKey` when the target path already existed:* `PathSummaryWriter#processFoundPathNode` (the merge branch of `adaptPathForChangedNode`, shared by XML renames and JSON subtree moves) adapted descendants and reference counts but never reset the moved node's own `pathNodeKey` — only the create-new-path branch did. The merge branch now resets the root like the create branch.

**3. REST API projection serving**
- Resource-scoped JSON queries are compiled with a session-bound vectorized executor pinned to the request's resolved revision (new `SirixCompileChain.createWithNodeAndJsonStore(nodeStore, jsonStore, session, revision)` factory, validated against the session's most recent revision), so the same analytical queries served from projections embedded are now served over HTTP.
- Brackit's analytical detection captures source paths, not resource identity, so the wiring sits behind a fail-closed `VectorizedServingGate` built as an **allowlist**: every `jn:doc` must name exactly the URL database/resource via two string literals; every other function call must be a known-safe unprefixed builtin or `xs:*` constructor — any prefixed function (`jn:open`, `jn:open-bitemporal`, `jn:open-revisions`, time-travel axes, `sdb:*`, `local:*`), unknown name, function reference (`name#arity`), or module import refuses the wiring. `nodeId`-scoped (subtree-bound) requests are excluded — the executor serves whole-resource projections and never reads the bound context item. A pinned non-latest revision additionally requires a pure context-item query (`jn:doc` always opens the most recent revision). Anything unprovable runs on the generic pipeline, so a refusal can only cost performance, never correctness.
- An adversarial self-review of the wiring (8 finder angles + per-candidate verification) confirmed and closed three gate-bypass paths before merge: `jn:open*` openers, function references, and `nodeId` subtree scoping; the gate's regression tests pin each one.

## Related issue

No issue — follow-up to #1116/#1117, closing the production-readiness gaps identified for the projection-index feature (soak coverage of the incremental-maintenance path and REST exposure).

## Type of change

- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Performance improvement
- [ ] Documentation only

## Checklist

- [x] `./gradlew build` passes locally (JDK 25) — all projection suites (core: 135 tests, query: 129 incl. the new stress suite and the 89-case differential suite), path-summary core suites, and the REST gate tests (18, offline task) run green; full sirix-query suite verified; the exhaustive sirix-core sweep is left to CI
- [x] Added or updated tests covering the change (`ProjectionIndexStressTest`, `VectorizedServingGateTest` incl. bypass regressions, a session-bound compile-chain serving test)
- [x] For behavioral changes to the storage engine: the `pathNodeKey` fix restores the documented path-summary attribution invariant (reference counts were already correct; node attribution now matches)
- [x] Updated documentation (README REST-serving section; `docs/KNOWN_LIMITATIONS.md` fixed-gaps entries for all three bugs)
- [x] No unrelated formatting churn (run Spotless / the `pre-commit` hook)

## Notes for reviewers

- The catalog-authoritative lookup change (`SirixVectorizedExecutor#lookupProjection`) intentionally demotes the in-memory registry to bench/test wiring when catalogued definitions exist — review the comment there for the staleness rationale. Known accepted edge: `hasProjections` is coarse (any catalogued def suppresses registry-only handles for other record sets) — perf-only, no live flow reaches it.
- `PathSummaryWriter#processFoundPathNode` is shared XML/JSON code; the one-line `resetPathNodeKey` addition mirrors the existing create-branch behavior (XML setName re-sets the same value afterwards, so it is a redundant-but-consistent write there).
- Known pre-existing (not introduced here, worth its own issue): a commit racing a resource-scoped GET can yield a response mixing revisions, because the context item/executor bind at request start while `jn:doc` opens the live most-recent at execution time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -763,7 +763,21 @@ readers keep their isolated snapshots. The full function
 family matches the other index types: `jn:find-projection-index($doc, $rootPath, $fields)` returns a
 projection's definition id (or `-1`), and `jn:drop-projection-index($doc[, $idx-no])` drops one or all
 projections (tombstoning the stored columns so a later same-shape re-creation rebuilds instead of
-serving leftovers). Current limits: column
+serving leftovers).
+
+Projection serving is also wired into the **REST API**: a resource-scoped query
+(`GET /database/resource?query=...`) is compiled with a vectorized executor bound to the request's
+resource and revision, so the same analytical queries are answered from the projection over HTTP.
+Because the analytical detection captures source paths — not resource identity — the REST layer
+applies a fail-closed serving gate: the executor is wired only when the query provably targets the
+request's own resource (every `jn:doc` names exactly that database/resource with two string
+literals; no other document/collection-opening function or module import appears; with a pinned
+non-latest revision, only pure context-item queries qualify). Anything unprovable simply runs on
+the generic pipeline — the gate can cost performance, never correctness. The index-management
+functions (`jn:create-projection-index`, `jn:find-projection-index`, `jn:drop-projection-index`)
+work over REST like any other JSONiq query.
+
+Current limits: column
 types are `long`, `boolean`, and `string` (floating-point columns are rejected rather than silently
 degraded); columns are resolved by trailing field name, which must be unique and unambiguous under the
 record set; queries that the projection cannot serve exactly (unrepresentable values, non-covered
--- a/README.md
+++ b/README.md
@@ -769,11 +769,13 @@ Projection serving is also wired into the **REST API**: a resource-scoped query
 (`GET /database/resource?query=...`) is compiled with a vectorized executor bound to the request's
 resource and revision, so the same analytical queries are answered from the projection over HTTP.
 Because the analytical detection captures source paths — not resource identity — the REST layer
-applies a fail-closed serving gate: the executor is wired only when the query provably targets the
-request's own resource (every `jn:doc` names exactly that database/resource with two string
-literals; no other document/collection-opening function or module import appears; with a pinned
-non-latest revision, only pure context-item queries qualify). Anything unprovable simply runs on
-the generic pipeline — the gate can cost performance, never correctness. The index-management
+applies a fail-closed serving gate built as an **allowlist**: the executor is wired only when the
+query provably targets the request's own resource — every `jn:doc` names exactly that
+database/resource with two string literals, every other function call is a known-safe builtin or
+`xs:*` constructor (any prefixed function, unknown name, function reference, or module import
+refuses), requests scoped to a `nodeId` subtree are excluded, and with a pinned non-latest
+revision only pure context-item queries qualify. Anything unprovable simply runs on the generic
+pipeline — the gate can cost performance, never correctness. The index-management
 functions (`jn:create-projection-index`, `jn:find-projection-index`, `jn:drop-projection-index`)
 work over REST like any other JSONiq query.
 
--- a/bundles/sirix-core/src/main/java/io/sirix/access/Databases.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/Databases.java
@@ -28,6 +28,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.UUID;
 
 import static java.util.Objects.requireNonNull;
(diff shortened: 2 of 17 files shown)


JohannesLichtenberger:
## Root-cause correction for commit `96157eb`

Commit `96157eb` is titled *"Fix silent no-op database removal poisoning re-created stores and query plans"* and frames its two changes as the fix for the flaky `OptimizerBehavioralE2ETest.casIndexTakesPrecedenceOverPath` failure. A follow-up investigation shows **that attribution was overstated**. Correcting the record here (I can't amend the commit message without a force-push, which would break #1127 stacked on this branch):

**What I actually verified**

1. **No handle leak exists.** I instrumented every `Database` open with its construction stack and ran the full 132-class `:sirix-query:test` suite. Across the entire module there is exactly **one** lingering-handle event, and its stack points at `OptimizerBehavioralE2ETest.restoredDatabaseWithLeakedHandleStaysExact` — the test that *deliberately* leaks a handle to prove the force-close path. No production code and no other test leaks a `json-path1` handle.
2. **Statistics-invalidation isn't demonstrably the cure either.** Full-module run with `StatisticsCatalog.invalidateDatabase(...)` disabled (force-close kept): green.
3. **Neither is the stress-test DB rename.** Full-module run with `ProjectionIndexStressTest` reverted to `json-path1`: green.

So none of the three candidate changes reproduces the flake when individually reverted locally. The `casIndexTakesPrecedenceOverPath` failure is **flaky / order-dependent** — it passed in isolation and in every recent full-module run, and I was **not** able to isolate its true trigger.

**What the two changes actually are**

They stand on their own as **defensive hardening of real latent bugs**, each proven by a targeted test — just not as the proven fix for that flake:

- `Databases.removeDatabase` now force-closes a still-registered handle instead of silently no-op'ing removal (which violated the documented "all sessions closed beforehand" contract). Proven by `restoredDatabaseWithLeakedHandleStaysExact`.
- `BasicJsonDBStore.removeIfExisting` clears the re-created database's cached histograms so the cost model can't be fed stale cross-generation statistics.
- `ProjectionIndexStressTest`'s dedicated DB names were made for a **separate, real** reason — on CI `:sirix-kotlin-api:test` runs in parallel and would wipe a shared `json-path1` mid-soak — not for this flake.

Net: the PR's substantive work is unaffected and CI is green; I just don't want the commit message to claim a root-cause fix I couldn't substantiate. If the flake resurfaces, it should be treated as an unresolved order-dependent test-isolation issue, not as regressed-by-this-change.

---

Issue #1125, opened 2026-07-19
Pass source-document identity into Brackit's VectorizedExecutor contract (upstream io.sirix:brackit fork)


JohannesLichtenberger:
## Summary

The canonical fix for the "vectorized executor can serve a query over the wrong resource" problem (see #1124) belongs in the **Brackit fork's extension contract**, not in any downstream gate or optimizer stage.

Today Brackit's `VectorizedGroupByDetection` captures a vectorizable pipeline's **source path** but not its **source document**. `VectorizedGroupByExpr` and the `VectorizedExecutor` SPI methods (`executeGroupByCount`, `executeAggregate`, `executeCountDistinct`, `executePredicateCount`, `executePredicateGroupByCount`, `executePredicateAggregate`, `executeGroupByCountMulti`, …) receive `String[] sourcePath` but **no document identity**. So a `SirixVectorizedExecutor` bound at construction to one `(session, revision)` has no way to verify that the pipeline it is being asked to answer actually targets that resource — it must trust the registration.

## Proposed change (in the Brackit fork)

1. In `VectorizedGroupByDetection`, capture the source `jn:doc('db','res'[, rev])` (or collection) identity of the matched `ForBind` alongside the existing `VECTORIZED_SOURCE_PATH_PREFIX`.
2. Carry that identity on `VectorizedGroupByExpr`.
3. Pass it to the `VectorizedExecutor` SPI methods (new parameter, or a small `SourceRef` value object), so the executor can compare it against its own binding at claim time and **decline** (return the "not handled" signal → generic pipeline) on a mismatch.

This closes the hole at the **contract level, for every caller**, with no downstream gate or AST-annotation-stripping stage required.

## Location / scope

The affected classes live in the **`io.sirix:brackit`** fork (a separate repository), e.g.:

- `io.brackit.query.compiler.optimizer.walker.topdown.VectorizedGroupByDetection`
- `io.brackit.query.expr.VectorizedGroupByExpr`
- `io.brackit.query.compiler.optimizer.VectorizedExecutor`
- `io.brackit.query.compiler.translator.SequentialPipelineStrategy` (`tryVectorizedExpr`)

Tracked from `sirixdb/sirix` because the consumer — `SirixVectorizedExecutor` (`bundles/sirix-query/.../scan/SirixVectorizedExecutor.java`) — and all the affected serving paths live here.

## Follow-through in sirix after the fork change ships

- Bump the `io.sirix:brackit` version in `sirix-query`.
[...]

PR #1127, opened 2026-07-19
Confine vectorized/projection serving to the bound resource in the optimizer (closes #1124)

### LINKED ISSUE #1124: Scope vectorized/projection serving to the bound resource in the optimizer (replace the REST textual gate)
## Problem

`SirixVectorizedExecutor` is constructed bound to a single `(session, revision)`, but Brackit's `VectorizedGroupByDetection` annotates a vectorizable scan by its **source path only** (`VECTORIZED_SOURCE_PATH_PREFIX` = field-name segments), never by **which document the FLWOR ranges over**. Brackit then applies the registered executor to any same-shaped pipeline in the compiled query — including one that reads a *different* database/resource. The executor cannot self-check, because it is never told the source document identity.

Left unguarded this returns **wrong data**: an executor bound to `sales.jn` answers an aggregate whose `for $r in jn:doc('db','other.jn')[] …` actually reads `other.jn`, from `sales.jn`'s projection/columns.

### Current mitigation (partial)

PR #1122 added a fail-closed **textual** guard, `VectorizedServingGate` (`bundles/sirix-rest-api/src/main/kotlin/io/sirix/rest/crud/json/VectorizedServingGate.kt`), that only wires the executor when the query text provably targets the request's own resource. Two gaps:

- It lives in the **REST layer** and lexes raw query text — the wrong altitude, and brittle.
- It only covers the **REST** caller. The embedded `SirixCompileChain.createWithJsonStore(store, session)` / `createWithNodeAndJsonStore(nodeStore, jsonStore, session, revision)` path has the identical blindness, guarded by javadoc only. An embedded caller running an aggregate over `jn:doc('db','B.jn')` through a chain bound to resource `A` still gets `A`'s projection answering `B`'s query.

## Proposed fix — a Sirix optimizer stage

Insert a new stage, `VectorizedResourceScopeStage`, **after** Brackit's `VectorizedGroupByDetection` (which the parent optimizer adds) and before translation. `SirixOptimizer` already exposes `addStageBeforeIndexMatching(...)` / `addStageAt(...)`, so the pipeline is Sirix-controlled.

The stage:

1. Receives the executor's bound identity `(database, resource, revision)`, threaded from `SirixCompileChain` (which already holds `autoExecutorSession` + `autoExecutorRevision`) into `SirixOptimizer`.
2. For every node carrying `VECTORIZED_SOURCE_PATH_PREFIX`, resolves the enclosing scan's **source document** — reusing `CostBasedStage.extractPathAndDocument()`, which already pulls `(db, res, revision)` from a `jn:doc`-rooted binding.
3. Decides:
   - **Context-item source** (`.[]`) → in-scope by construction (it *is* the bound rtx).
   - **`jn:doc(...)`** that does **not provably** equal the bound identity — different db/resource, or an explicit/mismatched revision (note: `jn:doc` opens the most-recent revision, so a request pinned to a non-latest revision must refuse a `jn:doc` scan) → **clear the `VECTORIZED_*` properties** on that node, so the translator builds the generic pipeline instead.

## Why this is the right altitude

- Operates on the **parsed AST**, not raw text.
- Covers **both** callers (embedded + REST) with one mechanism.
- Lets us **delete** the REST `VectorizedServingGate` entirely.

## Caveat to handle (fail-closed against drift)

The stage keys off Brackit's internal `VECTORIZED_*` property names. Brackit is version-pinned (`io.sirix:brackit:1.0-alpha7`). If a future Brackit bump renamed those keys, detection would set the new key, the translator would read the new key and still build the expr, but the stage would clear the old (absent) key → **fail open** → wrong-resource serving.

Mitigation: add an **end-to-end regression test** — a wrong-resource analytical query (bound to A, scans B) must return B's correct data via the generic pipeline, both embedded and over REST. Any Brackit bump that renames the annotation keys then fails the build loudly instead of silently failing open.

## Acceptance criteria

- [ ] A query whose vectorizable scan targets a resource other than the chain's bound resource returns correct results via the generic pipeline — verified for both the embedded API and REST.
- [ ] A request pinned to a non-latest revision does not serve a `jn:doc`-sourced analytical query from the pinned-revision executor.
- [ ] Same-resource analytical queries still serve from the projection/vectorized fast path (no regression to the #1122 serving tests).
- [ ] The REST `VectorizedServingGate` (and its 18-case test) is removed, its coverage migrated to the stage's tests.
- [ ] The drift regression test above is present.

## References

- PR #1122 (introduced the REST serving path + textual gate)
- `SirixOptimizer` (stage pipeline; `addStageBeforeIndexMatching`)
- `CostBasedStage.extractPathAndDocument()` (reusable `(db, res, revision)` extraction)
- `SirixCompileChain` (`autoExecutorSession`, `autoExecutorRevision`)
- Upstream contract fix that would supersede this: see the companion issue "Pass source-document identity into Brackit's VectorizedExecutor contract".

JohannesLichtenberger:
## What does this PR do?

Closes the "an executor bound to resource A can serve a query over resource B with A's data" hole at the right altitude — the query optimizer, on the parsed AST — replacing the interim REST textual gate from #1122 and covering the embedded caller too.

`SirixVectorizedExecutor` is bound to one `(session, revision)`, but Brackit's `VectorizedGroupByDetection` annotates a vectorizable scan by source **path** only, never by which document it dereferences. #1122 guarded only the REST caller with a fail-closed textual `VectorizedServingGate`; the embedded `createWithJsonStore(store, session)` path was guarded by javadoc only.

### Changes

- **`VectorizedResourceScopeStage`** (new) — runs after Brackit's detection (a parent-added optimizer stage) and before translation. Given a non-null `BoundResource` (i.e. the chain auto-wired an executor), it walks the AST and, if any `ForBind`'s scan source provably targets a resource other than the bound one — different database/resource, a revision the executor isn't pinned to (a bare `jn:doc` opens the most-recent revision), or an unprovable/dynamic `jn:doc` — strips the `VECTORIZED_*` annotations off every vectorized node so the translator builds the generic pipeline. Context-item scans and same-resource, same-revision `jn:doc` scans keep serving. **Fail-closed**: uncertainty → strip (only ever costs the fast path).
- **`BoundResource`** (new) — the `(database, resource, revision, revisionIsLatest)` the executor is bound to.
- **`SirixCompileChain`** resolves the `BoundResource` from its auto-executor session and threads it into `SirixOptimizer`; **`SirixOptimizer`** installs the stage before index matching when it is non-null.
- **Removed** the REST `VectorizedServingGate` and its 18-case test. `JsonGet` keeps only the `nodeId`-subtree refusal (which the stage does not model — the executor can't see a bound context item) and otherwise wires the session-bound chain unconditionally.
- **`VectorizedResourceScopeTest`** (new) proves: a bound-resource query still serves from its projection; a cross-resource query returns the OTHER resource's data (never the bound one's) and does not serve; a bare `jn:doc` under a non-latest binding does not serve.

## Related issue

Closes #1124. #1125 tracks the canonical upstream Brackit executor-contract change (pass the source-document identity into the `VectorizedExecutor` SPI so the executor self-verifies) that would subsume this in-repo stage.

## Type of change

- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Documentation only

## Checklist

- [x] `./gradlew build` — the new `VectorizedResourceScopeTest` and the #1122 projection serving/differential suites (`ProjectionIndexCatalogServingTest`, `ProjectionIndexStressTest`, `ProjectionIndexWtxServingTest`, the 89-case `TypedGroupByDifferentialTest`) pass; the full sweep is left to CI
- [x] Added tests covering the change
- [x] No unrelated formatting churn

## Notes for reviewers

- **Stacked on #1122** (base branch `claude/projection-index-readiness-tgpr44`) because it builds on that PR's session-bound `SirixCompileChain` infrastructure. Rebase onto `main` once #1122 merges.
- **Brackit annotation coupling** — the stage keys off Brackit's internal `VECTORIZED_*` property names (Brackit is pinned at `1.0-alpha7`). `VectorizedResourceScopeTest` is the drift guard: a Brackit bump that renamed those keys would make the cross-resource case return the wrong resource's data and fail the test loudly, rather than silently fail open. Note `getProperty(...) != null` is used to detect the marker, not `checkProperty(...)` — the marker's value is the source-path `String[]`, which `checkProperty` would `ClassCastException` on.
- The stage is a **no-op** when no executor was auto-wired (`bound == null`) — bench/test registrations via `SequentialPipelineStrategy.setVectorizedExecutor` and plain multi-resource chains are unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01QSTM7PdhRXfK1tCoT9LyeR)_


JohannesLichtenberger:
Superseded by the canonical fix and closing.

The upstream Brackit contract change (sirixdb/brackit#108) merged and published `io.sirix:brackit:1.0-alpha8-SNAPSHOT`, and #1128 (Closes #1125) is now merged to `main`. That implements `SirixVectorizedExecutor.acceptsSource`, which confines vectorized serving to the bound resource **at the executor-contract level** — the executor self-verifies the scan's `SourceRef` against its own bound `(database, resource, revision)`.

That subsumes this PR's `VectorizedResourceScopeStage` (plus the `BoundResource` threading and `SirixOptimizer` wiring): the resource decision now lives in one place — the executor — and needs no separate optimizer stage. So this interim in-repo approach is no longer needed.

Closing as superseded (not merged). The corresponding interim issue #1124 is likewise subsumed by #1125.

---

PR #1156, opened 2026-07-28
Make the prefetcher close()-drain test deterministic

JohannesLichtenberger:
## What does this PR do?

Fixes the CI failure in [run 30334189487](https://github.com/sirixdb/sirix/actions/runs/30334189487), where `Test sirix-core` failed with:

```
RevisionPrefetcherLifecycleTest > close_doesNotReturnWhileInFlightTasksStillHoldTransactions
java.lang.AssertionError: a prefetch task must park inside the open, holding an rtx
```

The failure was on the harness's setup latch, not on the contract assertion — the test never reached the state it wanted to measure.

**Root cause — a race in the test, not in `RevisionPrefetcher`.** The consumer thread called `poll()` and `close()` back to back. `close()` publishes `closed = true`, and every prefetch body checks that flag *before* calling `beginNodeReadOnlyTrx` (cancellation checkpoint #1, `RevisionPrefetcher.java:158`). The test document has three revisions, so revisions 2 and 3 are the only tasks that park. When their virtual threads had not been scheduled by the time `close()` published the flag, both short-circuited without ever opening a trx — nothing parked, and `parked.await(20s)` expired. That window is microseconds of consumer work against thread scheduling on a runner with 12 parallel jobs, which is why it passes locally and flakes in CI.

**The fix** adds the handshake the test's own javadoc already assumed ("a proxy session opens the REAL rtx and then parks … so the task is provably holding an open, counted transaction while `close()` runs"):

- The consumer waits on a new `closeAllowed` latch before calling `close()`; the harness releases it only once a task has actually parked.
- The harness waits for `prefetcher.isClosed()` before releasing the parked tasks, so the release can never precede the flag. This also replaces the previous "sleep 300ms and hope the consumer got from `poll()` to `close()`" assumption with an observed state; the remaining 300ms sleep now only gives a fire-and-forget `close()` time to record its above-baseline count.
- Consumer-thread exceptions are captured and asserted instead of surfacing only as the `-1` sentinel count.

Production code is unchanged — the diff is one test file.

## Related issue

No issue; this is a direct fix for the failing scheduled CI run linked above.

## Type of change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Performance improvement
- [ ] Documentation only

## Checklist

- [x] Added or updated tests covering the change (the change *is* the test)
- [x] No unrelated formatting churn
- [ ] `./gradlew build` passes locally (JDK 25) — ran the affected module's test task, `:sirix-core:test --tests io.sirix.axis.temporal.RevisionPrefetcherLifecycleTest`, rather than a full build; leaving the full run to CI
- [ ] For behavioral changes to the storage engine: n/a — no production code touched
- [ ] Updated documentation — n/a

## Notes for reviewers

Verified in both directions, since a test that only ever passes proves nothing:

- As committed: green.
- With `close()` temporarily reverted to fire-and-forget (early return before the drain join): fails on the contract assertion — `expected:<1> but was:<3>`, the baseline plus the two parked rtxs — not on a latch timeout. So the test still pins the synchronous-release contract from `RevisionPrefetcher.close()`, and now fails for the right reason when that contract breaks.

The three latch waits share one `HANDSHAKE_TIMEOUT_SECONDS = 20` bound. Each guards a step that takes microseconds when it happens at all, so it is a deadlock backstop sized for a saturated runner, not a timing assumption the test depends on.

---
_Generated by [Claude Code](https://claude.ai/code/session_0144wbptK412oPDuNdNv2gr3)_

### DIFF
--- a/bundles/sirix-core/src/test/java/io/sirix/axis/temporal/RevisionPrefetcherLifecycleTest.java
+++ b/bundles/sirix-core/src/test/java/io/sirix/axis/temporal/RevisionPrefetcherLifecycleTest.java
@@ -16,6 +16,7 @@ import java.lang.reflect.Proxy;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.IntSupplier;
 
 import static org.junit.Assert.assertEquals;
@@ -42,6 +43,13 @@ import static org.junit.Assert.fail;
  * </ul>
  */
 public final class RevisionPrefetcherLifecycleTest {
+
+  /**
+   * Bound on every handshake in {@link #close_doesNotReturnWhileInFlightTasksStillHoldTransactions}.
+   * Each one waits for a step that takes microseconds when it happens at all, so this is a
+   * deadlock backstop generous enough for a saturated CI runner, not a tuning knob.
+   */
+  private static final long HANDSHAKE_TIMEOUT_SECONDS = 20L;
 
   private Holder holder;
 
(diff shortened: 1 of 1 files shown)


JohannesLichtenberger:
Closing as incorporated — the approach here was adopted in #1157.

We hit this flake from two directions at once: this PR came from the scheduled CI failure on `main`, while #1154 reached the same root cause while unblocking a batch of Dependabot updates. #1154 merged first, so `main` briefly carried the weaker of the two fixes.

The version here is better on one specific point, which is why it is the one that survived. #1154 relied on a 300 ms sleep for the consumer to get from the parked latch to `close()`. The failure mode of that sleep being too short is not a red test — it is a green one that has silently stopped discriminating: if the parked tasks resume before `close()` runs, they complete normally, `close()` drains them through the completion callback, and `activeTrxCount` returns to the baseline whether or not `close()` is synchronous. Waiting on `isClosed()` instead, as this PR does, makes the ordering observed rather than assumed and removes that failure mode.

#1157 carries that change, plus the consumer-thread exception capture and the shared `HANDSHAKE_TIMEOUT_SECONDS` bound, rebased onto `main` after #1154. It is verified in both directions there: green as committed, and still failing on the contract assertion (`expected:<1> but was:<3>`) against a temporary fire-and-forget `close()`, so it continues to pin the synchronous-release contract.

No action needed here; the analysis in this PR's description is the clearer write-up of the race and is worth keeping for reference.

---

PR #1160, opened 2026-08-02
feat(sirix-core): add an exact-decimal column and answer predicate trees as selection bitmaps

JohannesLichtenberger:
## Intent

GOAL. Close the ALP gap so real decimals are scannable from columns, then fix the architecture that made that gap expensive to close in the first place. Plus three defects an earlier code review found.

STRAND 1 - an exact decimal column (ENC_DEC). A decimal joined the double column only when its double image was EXACT, which only dyadic rationals are, so 19.99 / 100.10 and essentially every real price fell back to the record heap and the ALP path was unreachable for the data it exists for. ENC_DEC carries such values as their own unscaled integers at a common scale, reusing the ALP block layout with e = scale, f = 0 and no exceptions.

DELIBERATE and load-bearing: the kernels REFUSE a double bound on an ENC_DEC tag. Two decimals can share a double image (1000.25000000000001 and 1000.25 do), so answering one from the other silently puts rows on the wrong side. An earlier attempt in this same session changed ONLY the storage side, left the bound in double space, and returned 0 where the answer is 2000; it was reverted. DecimalDoubleCollisionTest pins that invariant. The bound is therefore converted in DECIMAL space by integer arithmetic at the tag's own scale (decBoundAtScale), which is what makes the comparison exact rather than merely close. Refusal is ENUMERATED rather than a default arm: an unrecognised encoding refuses instead of falling into countPlain, so a path nobody wired is slow, never wrong. All four kernel call sites check the REFUSED sentinel; an unchecked one would be a catastrophic miscount.

STRAND 2 - predicate trees as selection bitmaps. A kernel returning a scalar count cannot be composed (two counts cannot be ANDed), so every conjunction had to be fused BY HAND, once per encoding and again for the liveness-masked variant. Leaves now answer as bitmaps in their own encoded domains and And/Or/Not are word operations. Nothing decodes except ENC_ALP_RD, whose bit split is not order-preserving - that is a property of the SCHEME, not a gap.

DELIBERATE: RegionCountPlan is a FLATTENED predicate (a closed union of shapes) and flattening is lossy, so the tree had to be threaded back to the scan (CompiledPredicate.tree -> TreeRoute -> withTreeRoute, resolved once per scan) BEFORE the evaluator could be wired anywhere. The tree route is PRIMARY for multi-leaf conjunctions; the hand-fused kernel is deliberately KEPT as a fallback for pages the tree declines (it needs record-ordinal linkage and refuses a non-zero alignment lead). Tier-2 kernels (zone-map-settled, single-interval count) are deliberately KEPT: they do strictly less work than bitmap+popcount, which is the whole BtrBlocks/Umbra advantage over decode-and-filter. Two selection currencies coexist on purpose - index vectors after a filter (what DuckDB/Umbra push between operators), bitmaps inside one.

STRAND 3 - three review defects: PrefetchingChildAxis reset ran from AbstractAxis's constructor against a half-built object and called an overridable close(); RecordPagePrefetcher dropped the teardown handoff on its RejectedExecutionException path, stranding worker transactions; RevisionTrackingExecutor.canExecute released a pin it could not prove it owned, because nested evaluations share a QueryContext - a wrong revision is worse than a bounded retention.

TESTING approach, deliberately: popcount(select) == count asserted PER ENCODING against brute force, because a select that is merely close is invisible to count-based tests but puts rows on the wrong side of every AND/OR. The algebra checked against a brute-force interpreter over 500 random trees. PredicateTreeRouteTest asserts the route is actually REACHED, because a fallback validated only by "it agrees" passes vacuously while it is dead code - it WAS dead until it became primary. Negative controls were run on the ALP fix (revert only src/main, confirm the tests fail). 12,470 tests green across both repos.

KNOWN and intended: a threshold whose own double image is inexact is served in decimal space; multi-interval and versioned-merge arms serve decimal tags; a shape none covers falls back - correct, not accelerated.

CROSS-REPO: SirixOptimizer opts into a new brackit capability (per-INSTANCE constructor parameter, deliberately not a static flag, which would latch on class initialization and affect a plain CompileChain sharing the JVM). That brackit change is sirixdb/brackit#113 and is NOT yet published, so this branch cannot compile against the published 1.0-alpha10-SNAPSHOT - it builds against a locally installed brackit. A compile failure on that constructor is expected until #113 lands, not a defect in this diff.

RERUN CONTEXT. A previous run (01KZEYHWJNP0SM4TCBMBMCY8V0) cleared review in three rounds and then FAILED at the test step with "claude exited: exit status 1" and an empty message - an agent-process exit, not a failing test: that agent's own log ends "Everything is validated. Final state check is clean." Its two fix commits are now the branch head (89a97b3cc, 25c208dea) and were verified locally before this rerun: sirix-query 1111 tests and sirix-core page/node/io/cache/access 7269 tests, 0 failures. Those commits also ADDED tests - DecimalBoundSaturationTest, TreeRouteRegressionTest, UnsoundDecimalTagScopeTest, DoubleRegionUnsoundTagScopeTest - plus mixed-corpus cases in DecimalDoubleCollisionTest and versioned-merge cases in VersioningColumnScanTest. Treat those as intended, not as stray files.

## What Changed

- **Exact-decimal column (`ENC_DEC`).** Decimals previously joined the double column only when their double image was exact — true only for dyadic rationals — so ordinary values like `19.99` fell back to the record heap and the ALP path was unreachable for the data it exists for. `DoubleRegion` now carries such values as their own unscaled integers at a common scale, reusing the ALP block layout (`e = scale`, `f = 0`, no exceptions). The kernels deliberately **refuse** a double bound on an `ENC_DEC` tag — two distinct decimals can share one double image, so answering from the rounded bound silently puts rows on the wrong side — and convert thresholds in decimal space by integer arithmetic at the tag's own scale. Refusal is enumerated per encoding rather than a default arm, so an unwired encoding is slow, never wrong, and all four kernel call sites check the `REFUSED` sentinel.
- **Predicate trees as bitmaps.** A kernel returning a scalar count cannot be composed, so every conjunction had to be hand-fused once per encoding. `PredicateBitmapEvaluator` lets leaves answer as selection bitmaps in their own encoded domains, making `And`/`Or`/`Not` word operations; nothing decodes except `ENC_ALP_RD`, whose bit split is not order-preserving. Because `RegionCountPlan` is a lossy flattening, the tree is threaded back to the scan (`CompiledPredicate.tree` → `TreeRoute` → `withTreeRoute`, resolved once per scan) and is primary for multi-leaf conjunctions; the hand-fused kernel stays as the fallback for pages the route declines, and the tier-2 zone-map and single-interval kernels stay because they do strictly less work than bitmap+popcount.
- **Review defects and supporting work.** Fixed `PrefetchingChildAxis` resetting from `AbstractAxis`'s constructor against a half-built object via an overridable `close()`, `RecordPagePrefetcher` dropping the teardown handoff on its `RejectedExecutionException` path (stranding worker transactions), and `RevisionTrackingExecutor.canExecute` releasing a pin it could not prove it owned. Alongside: SIMD kernels for the number/string/boolean/double regions, new record-ordinal and zone-map regions, path-summary child indexing and statistics corrections, batched page prefetching in the reader stack, the JMH benchmark set with the PostgreSQL bulk comparison, and deduplication of the architecture report onto the `.tex` original.

*Build note:* `SirixOptimizer` opts into a new brackit capability through a per-instance constructor parameter (deliberately not a static flag, which would latch on class initialization and affect a plain `CompileChain` sharing the JVM). That change is sirixdb/brackit#113 and is not yet published, so this branch builds against a locally installed brackit; a compile failure on that constructor is expected until #113 lands.

## Risk Assessment

⚠️ Medium: Both round-1 error findings are correctly fixed with a further soundness hardening that closes the same rounded-threshold hole inside the tree route, and I found no new defects in the fix commit; the residual risk is the branch's own size and reach — a new on-disk double-region encoding plus a rewritten scan hot path — rather than any substantiated problem in the code.

## Testing

Exercised the three strands with targeted suites in both modules — the decimal-column and kernel-soundness tests, the bitmap-algebra and tree-route tests (including the assertion that the route is actually reached), and the prefetch/executor-pin tests — all green, and confirmed the branch compiles against the locally installed brackit. Beyond the suites I drove the real end-user surface: 20,000 orders with genuine two-decimal prices shredded via jn:store, then queried with ordinary JSONiq; every decimal predicate returned the same count as the record heap and an independent BigDecimal oracle while being served entirely from columns (118 pages, 0 fallbacks), composed shapes routed 79 and 118 pages through the predicate tree, the 1000.25 / 1000.25000000000001 corpus answered 1000 on both paths where a double-space bound answers 0, and a projection query returned the actual rows. A negative control that disabled the exact-decimal column flipped five committed assertions to served=0/fellBack=79, so the column-path evidence is load-bearing rather than vacuous. No screenshot or rendered-HTML artifact applies: this change is a storage/query-engine path with no UI, so the reviewer-visible surface is the JSONiq query transcript. One shape (a decimal-and-integer conjunction) is correct but not accelerated, which the intent names as known and intended.

<details>
<summary>Evidence: JSONiq query transcript — real decimal prices served from columns</summary>

<code>count(for $u in jn:doc(&#39;evidence-orders-db&#39;,&#39;orders.jn&#39;)[] where $u.price gt 19.99 return $u)&#10;BigDecimal oracle : 20000&#10;record heap : 20000&#10;typed columns : 20000&#10;pages from columns : 118&#10;pages -&gt; record heap : 0&#10;&#10;count(... where $u.price ge 100.10 and $u.price le 200.90 ...) oracle 4040 | records 4040 | columns 4040 | 118 pages from columns, 0 fallbacks&#10;count(... where $u.price lt 19.99 or $u.price gt 500.05 ...) oracle 2000 | records 2000 | columns 2000 | 118 pages from columns, 0 fallbacks&#10;count(... where $u.price eq 100.50 ...) oracle 40 | records 40 | columns 40 | 118 pages from columns, 0 fallbacks&#10;&#10;count(... where $u.year gt 1950 and not($u.active) ...) records 3918 | columns 3918 | 79 pages via predicate tree&#10;count(... where $u.title eq &#34;Beta&#34; or $u.year lt 1910 ...) records 7747 | columns 7747 | 118 pages via predicate tree&#10;&#10;SOUNDNESS - two decimals that share one double image&#10;corpus: 2,000 rows alternating 1000.25 and 1000.25000000000001&#10;(double) 1000.25000000000001 &gt; (double) 1000.25 -&gt; false [a double-space bound counts 0]&#10;BigDecimal compare &gt; 0 -&gt; true [decimal space counts 1000]&#10;count(... where $u.price gt 1000.25 ...) record heap : 1000 columns : 1000&#10;&#10;RESULT SET - the rows themselves, off the column path&#10;for $u in jn:doc(...)[] where $u.price ge 549.99 and $u.id lt 1500 return {&#34;id&#34;: $u.id, &#34;price&#34;: $u.price}&#10;{&#34;id&#34;:499,&#34;price&#34;:549.99}&#10;{&#34;id&#34;:999,&#34;price&#34;:549.99}&#10;{&#34;id&#34;:1499,&#34;price&#34;:549.99}</code>

```text
=========================================================================
 SirixDB - JSONiq over a real price field (20,000 orders, 2 decimals)
 Each query is answered twice: once off the record heap (ground truth),
 once off the typed columns. Counters show which machinery served it.
=========================================================================

jn:store('evidence-orders-db','orders.jn', [ {"id":0,"price":50.00,"year":1900,...},
                                       {"id":1,"price":51.01,"year":1901,...}, ... ])


-------------------------------------------------------------------------
 STRAND 1 - decimal predicates served from columns (ENC_DEC)
-------------------------------------------------------------------------

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.price gt 19.99 return $u)
      BigDecimal oracle    : 20000
      record heap          : 20000
      typed columns        : 20000
      pages from columns   : 118
      pages -> record heap : 0

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.price ge 100.10 and $u.price le 200.90 return $u)
      BigDecimal oracle    : 4040
      record heap          : 4040
      typed columns        : 4040
      pages from columns   : 118
      pages -> record heap : 0

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.price lt 19.99 or $u.price gt 500.05 return $u)
      BigDecimal oracle    : 2000
      record heap          : 2000
      typed columns        : 2000
      pages from columns   : 118
      pages -> record heap : 0

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.price eq 100.50 return $u)
      BigDecimal oracle    : 40
      record heap          : 40
      typed columns        : 40
      pages from columns   : 118
      pages -> record heap : 0

-------------------------------------------------------------------------
 STRAND 2 - composed predicates, answered as bitmap algebra
-------------------------------------------------------------------------

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.price gt 300.33 and $u.year gt 1950 return $u)
      BigDecimal oracle    : 5877
      record heap          : 5877
      typed columns        : 5877
      pages from columns   : 0
      pages via pred. tree : 0
      pages -> record heap : 0 (columns read, shape refused)
      pages w/o column path: 0

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where not($u.price gt 300.33) return $u)
      BigDecimal oracle    : 10000
      record heap          : 10000
      typed columns        : 10000
      pages from columns   : 118
      pages via pred. tree : 0
      pages -> record heap : 0 (columns read, shape refused)
      pages w/o column path: 0

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.year gt 1950 and not($u.active) return $u)
      record heap          : 3918
      typed columns        : 3918
      pages from columns   : 79
      pages via pred. tree : 79
      pages -> record heap : 39 (columns read, shape refused)
      pages w/o column path: 0

  count(for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.title eq "Beta" or $u.year lt 1910 return $u)
      record heap          : 7747
      typed columns        : 7747
      pages from columns   : 354
      pages via pred. tree : 118
      pages -> record heap : 0 (columns read, shape refused)
      pages w/o column path: 0

-------------------------------------------------------------------------
 SOUNDNESS - two decimals that share one double image
-------------------------------------------------------------------------
  corpus: 2,000 rows alternating 1000.25 and 1000.25000000000001
  both round to the SAME double; only exact decimal comparison separates them

  in Java, for reference:
      (double) 1000.25000000000001 > (double) 1000.25   ->  false   [a double-space bound counts 0]
      new BigDecimal("1000.25000000000001").compareTo(new BigDecimal("1000.25")) > 0  ->  true   [decimal space counts 1000]

  count(... where $u.price gt 1000.25 ...)
      record heap : 1000
      columns     : 1000
      verdict     : OK - both answer 1000 (the odd rows); a double-space bound answers 0

-------------------------------------------------------------------------
 RESULT SET - the rows themselves, off the column path
-------------------------------------------------------------------------
  for $u in jn:doc('evidence-orders-db','orders.jn')[] where $u.price ge 549.99 and $u.id lt 1500 return {"id": $u.id, "price": $u.price}

      {"id":499,"price":549.99}
      {"id":999,"price":549.99}
      {"id":1499,"price":549.99}

=========================================================================
 All column-path answers matched the record heap and an independent
 BigDecimal oracle; every decimal query above was served from columns.
=========================================================================
```
</details>
<details>
<summary>Evidence: Negative control — exact-decimal column disabled (src/main patch, reverted)</summary>

<code>patch: DoubleRegion.encodeTag &#39;if (decE &gt;= 0)&#39; -&gt; &#39;if (false)&#39;&#10;&#10;FAILED-TEST: AllDoubleColumnScanTest &gt; inexactThresholdIsServedFromTheDecimalColumn(): no page served from the typed columns (served=0, fellBack=79, unavailable=0) for $u.price gt 19.99 ==&gt; expected: &lt;true&gt; but was: &lt;false&gt;&#10;FAILED-TEST: AllDoubleColumnScanTest &gt; allDoubleFieldIsServedFromTheDoubleColumn(): served=0, fellBack=79 for $u.price gt 100.25&#10;FAILED-TEST: AllDoubleColumnScanTest &gt; disjunctionOverFractionalFieldIsServed(): served=0, fellBack=79&#10;FAILED-TEST: AllDoubleColumnScanTest &gt; conjunctionInsideDisjunctionIsServed(): served=0, fellBack=79&#10;FAILED-TEST: AllDoubleColumnScanTest &gt; halfDoubleFieldIsServedFromBothColumns(): served=0, fellBack=79&#10;&#10;(answers stayed correct throughout — the record heap still compares decimals exactly; what died was the column path, which is exactly what these assertions measure)</code>

```text
NEGATIVE CONTROL - exact-decimal column (ENC_DEC) disabled in DoubleRegion.encodeTag
  patch: 'if (decE >= 0)' -> 'if (false)'  (src/main only, reverted afterwards)

===== FAILED TESTS (6) =====
FAILED-TEST: io.sirix.query.scan.AllDoubleColumnScanTest > conjunctionInsideDisjunctionIsServed(): org.opentest4j.AssertionFailedError: no page served from the typed columns (served=0, fellBack=79, unavailable=0) for ($u.price ge 100.05 and $u.price le 200.05) or $u.price gt 500.05 ==> expected: <true> but was: <false>
FAILED-TEST: io.sirix.query.scan.AllDoubleColumnScanTest > allDoubleFieldIsServedFromTheDoubleColumn(): org.opentest4j.AssertionFailedError: no page served from the typed columns (served=0, fellBack=79, unavailable=0) for $u.price gt 100.25 ==> expected: <true> but was: <false>
FAILED-TEST: io.sirix.query.scan.AllDoubleColumnScanTest > inexactThresholdIsServedFromTheDecimalColumn(): org.opentest4j.AssertionFailedError: no page served from the typed columns (served=0, fellBack=79, unavailable=0) for $u.price gt 19.99 ==> expected: <true> but was: <false>
FAILED-TEST: io.sirix.query.scan.AllDoubleColumnScanTest > disjunctionOverFractionalFieldIsServed(): org.opentest4j.AssertionFailedError: no page served from the typed columns (served=0, fellBack=79, unavailable=0) for $u.price lt 19.99 or $u.price gt 500.05 ==> expected: <true> but was: <false>
FAILED-TEST: io.sirix.query.scan.AllDoubleColumnScanTest > halfDoubleFieldIsServedFromBothColumns(): org.opentest4j.AssertionFailedError: no page served from the typed columns (served=0, fellBack=79, unavailable=0) for $u.mixed gt 100.25 ==> expected: <true> but was: <false>
FAILED-TEST: io.sirix.query.scan.DecimalColumnEvidenceTranscriptTest > transcript(): org.opentest4j.AssertionFailedError: no page served from columns: $u.price gt 19.99 ==> expected: <true> but was: <false>
===== END FAILED TESTS =====

NOTE: DecimalColumnEvidenceTranscriptTest was a TRANSIENT harness used only to produce
the transcript artifact; it was deleted after the run. The five AllDoubleColumnScanTest
failures above are committed tests, and they are the load-bearing part of this control.
```
</details>

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 1 info</summary>

- 🚨 `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:11064` - foldBranchInterval discards the doublesServable=false it just computed. Line 11017 sets the local flag false when a DEC_CMP literal has no faithful double image (mirroring foldNumericInterval:11397, which is the deliberate guard against answering a decimal threshold in double space), but line 11064 returns `new BranchInterval(lo, hi, dlo, dhi, true, ...)` with the flag hardcoded. planNumericDisjunction (line 10837) ANDs iv.doublesServable() into its own flag, so the disjunction&#39;s orDblIntervals are built from the rounded image AND marked servable — the exact substitution the single-interval path refuses to make. Failure: field `p` on a page carries both a genuinely xs:double value and a BigDecimal value equal to 19.989999999999998436805981327779591083526611328125 (the double image of 19.99). decimalImagesAreExact() admits that mixed tag to ENC_ALP over double images. Query `p lt 19.99 or p gt 100.10` folds branch 1 to dhi = nextDown(image(19.99)), so the stored value is not counted; the record path evaluates it as FK_DECIMAL via evalDecOnDecimal (19.98999...125 &lt; 19.99 = true) and counts it. Silent undercount. Before this diff the same case returned null from foldBranchInterval and took the record path. Fix: return `doublesServable` instead of `true` at line 11064, matching foldNumericInterval.
- 🚨 `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:13067` - The last-resort tree route breaks the invariant the null patch scratch relies on. countPageViaTree sets scratch.pendingFusedAnchorSlot (line 11678) whenever anchorLead == 1, and the new call at line 13067 runs it for ANY plan kind — not just fused. The batch arm allocates the patch scratch only for fused plans (`patchScratch = fusedPlan ? new EvalScratch(...) : null`, line 5932, whose comment says &#34;the scratch exists only when a fused plan can hand one back&#34;). columnarPageCount then calls patchBoundaryAnchor unconditionally (line 12997), which reaches `loadFields(rtx, cp, evalScratch)` with a null scratch and dereferences `scratch.fieldKind.length` at line 6061. Failure: single-field predicate `$x.v gt 0.000000000000001` (scale 15 &gt; DoubleRegion.MAX_DECIMAL_SCALE, so decIntervalOf returns null; the image is inexact, so doublesServable is false) over an all-double ENC_ALP field. countPageCombiningDoubles refuses at line 13272, c &lt; 0, and the tree route serves the page from double bounds — but if that page&#39;s first anchor occurrence belongs to a record spanning in from the previous page (anchorLead == 1) the pending slot is handed back and the worker thread throws NullPointerException. BATCH_GENERIC_EVAL_ENABLED defaults to true, so this is the default arm. Fix: allocate patchScratch whenever regionPlan != null (or whenever regionPlan.treeRoute() != null).
- ℹ️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:13062` - For a fused plan the tree route is evaluated twice per page whenever both it and the fused kernel decline: once inside countPageFromRegions (line 11559) and again at line 13067 after c &lt; 0. The inputs are identical and the routine is deterministic, so the second run always declines too — it re-decodes the record ordinals, re-scans the name-key bitmap and re-walks the tree for nothing, on the fallback path of every page of such a scan. Separately, when the first attempt returned NEED_STRING_DICTIONARY and the retry fetched `withDict`, line 13067 still passes `regions` (the dictionary-less page), so a fused plan with a string leaf can never be served by the last-resort route even though the dictionary is now in hand. Skip the second call when plan.isFused() (it already ran), and pass the dictionary-bearing page when one was fetched.
- ℹ️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:10531` - attachTreeRoute returns the plan unchanged when it is null, and columnarPageCount is only entered when regionPlan != null, so the tree route can never serve a shape planRegionCount could not flatten. A multi-field disjunction such as `$u.year gt 1990 or $u.active` makes planFusedCount return null at line 10631 (top level is not AND), so no plan exists and the record path is taken without the evaluator ever running. This is consistent with the intent&#39;s &#34;a shape none covers falls back - correct, not accelerated&#34;, but the comments at lines 11621-11627 and 13063-13066 (&#34;the reason a shape nobody wrote a kernel for no longer costs a record reconstruction&#34;) describe a reach the wiring does not have; the route&#39;s actual scope is fused conjunctions plus pages a shape kernel declined. Flagged so the comments are not read as a coverage guarantee later.
- ℹ️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:13387` - countDecUnion allocates `new long[branches.length * 2]` per page, and mergeIntervals (line 11091) allocates its own output plus a possible Arrays.copyOf, so a disjunction over an ENC_DEC field allocates two or three arrays per page for the whole scan. The double union is merged once at plan time precisely to avoid this; the decimal union cannot be, because the inclusive bounds depend on the tag&#39;s scale — but the scale is drawn from a small fixed range, so the converted bounds could be memoised per scale in the RegionHeaderScratch (which already carries per-worker buffers) instead of rebuilt per page. Minor in absolute terms, but it is GC pressure on the hot scan loop the project&#39;s HFT rules single out.

🔧 Fix: Keep decimal thresholds out of double space, restore patch scratch
1 info still open:
- ℹ️ `bundles/sirix-query/src/test/java/io/sirix/query/scan/TreeRouteNonFusedSeamTest.java:142` - The regression test added for the patch-scratch fix does not pin the path it protects. theRouteDeclinesAnUnrepresentableThreshold asserts regionTreePages() == 0 for all four NON_FUSED_SHAPES — correct, because the new imageFaithful guard in selectDouble (line 12054) refuses exactly those scale-15 thresholds on an ALP-over-doubles tag — so no page is ever served through the last-resort route in this class. The other two tests (nonFusedTreeRoutePagesAgreeAcrossSeams, repeatScanAgrees) only compare the column path against the record path, which agrees for free on a page nobody serves columnar. The restored invariant (a non-fused plan&#39;s route hands back a spanning anchor slot and patchBoundaryAnchor settles it against a now-allocated EvalScratch) is therefore exercised by no assertion, which is the vacuous-fallback failure mode the intent&#39;s own testing section calls out (&#34;a fallback validated only by &#39;it agrees&#39; passes vacuously while it is dead code&#34;). The fix itself is unconditionally correct — patchScratch is allocated whenever regionPlan != null (line 5933) and the fallback path clears the pending slot (line 13117) — so this is a coverage note, not a defect. Reaching the route on a non-fused plan needs a page that carries KIND_RECORD_ORDINAL even though only REGION_FUSED_COUNT_MASK requests it (a resident page carrying a superset), plus a leaf the route can answer; repeatScanAgrees already constructs the residency half of that setup, so pinning regionTreePages() &gt; 0 for a shape with an exactly-representable threshold would close it.

</details>

<details>
<summary>✅ **Test** - passed</summary>

✅ No issues found.
- `./gradlew :sirix-query:test --tests io.sirix.query.scan.DecimalDoubleCollisionTest --tests io.sirix.query.scan.DecimalBoundSaturationTest --tests io.sirix.query.scan.UnsoundDecimalTagScopeTest --tests io.sirix.query.scan.DecimalDisjunctionSoundnessTest --tests io.sirix.query.scan.AllDoubleColumnScanTest`
- `./gradlew :sirix-query:test --tests io.sirix.query.scan.PredicateBitmapEvaluatorTest --tests io.sirix.query.scan.PredicateTreeRouteTest --tests io.sirix.query.scan.TreeRouteRegressionTest --tests io.sirix.query.scan.TreeRouteNonFusedSeamTest --tests io.sirix.query.scan.PredicateTreeCountIntegrationTest --tests io.sirix.query.scan.VersioningColumnScanTest --tests io.sirix.query.scan.RevisionTrackingExecutorTest --tests io.sirix.query.json.JsonDBArrayScanTeardownTest`
- `./gradlew :sirix-query:test --tests io.sirix.query.scan.RegionOnlyPredicateCountTest --tests io.sirix.query.json.JsonDBArrayMutationMemoTest --tests io.sirix.query.AutoWiredExecutorTest`
- `./gradlew :sirix-core:test --tests io.sirix.page.pax.DoubleRegionUnsoundTagScopeTest --tests io.sirix.page.pax.ColumnKernelEquivalenceTest`
- `./gradlew :sirix-core:test --tests io.sirix.page.pax.DoubleRegionSelectEquivalenceTest --tests io.sirix.page.pax.LeafSelectEquivalenceTest --tests io.sirix.page.pax.DoubleRegionAlpTest`
- <code>Manual end-to-end evidence run: shredded 20,000 orders with two-decimal prices via `jn:store`, then executed the decimal, composed and collision JSONiq queries through `SirixVectorizedExecutor` with the column path both off and on, recording `regionOnlyPagesServed` / `regionOnlyPageFallbacks` / `regionTreePages` and a projection result set (transcript artifact)</code>
- <code>Negative control: patched `DoubleRegion.encodeTag` (`if (decE &gt;= 0)` -&gt; `if (false)`) to disable the exact-decimal column, confirmed 5 committed `AllDoubleColumnScanTest` assertions flip to `served=0, fellBack=79`, then restored the file with `git checkout --`</code>
- <code>`./gradlew :sirix-query:compileTestJava` (verifies the branch builds against the locally installed brackit despite the unpublished #113 constructor)</code>

</details>

<details>
<summary>⚠️ **Document** - 1 info</summary>

- ⚠️ `docs/SIRIX_ARCHITECTURE_REPORT.md:497` - docs/SIRIX_ARCHITECTURE_REPORT.md (added on this branch, commit b6fa64475) is a near-duplicate of the pre-existing docs/SIRIX_ARCHITECTURE_REPORT.tex — §8 &#34;Vectorised and Columnar Execution&#34; states the same facts in both, in places verbatim. That section is now stale in BOTH copies: it describes the executor as extracting &#34;conjunctive per-column predicates (fusing &gt; + &lt; into one BETWEEN)&#34; as the mechanism, which this change demoted to a fallback behind the predicate-tree bitmap evaluator, and §8.2 describes PAX regions as numeric/boolean/string only, with no typed double column or exact-decimal encoding. I deliberately did not edit either copy: fixing one leaves the other wrong, and updating both is the prose-synchronisation the placement policy forbids. Follow-up worth deciding: pick one owner (the .tex is the original and predates the branch) and reduce the other to a pointer, or generate the .md from the .tex.
- ℹ️ `docs/COMPARISON_POSTGRES_BULK.md:401` - docs/COMPARISON_POSTGRES_BULK.md §4.1 lists &#34;Prefetch the chain one hop ahead&#34; as recommendation #2 of what the measurements point at — work this change implemented (PrefetchingChildAxis / RecordPagePrefetcher). I left it as-is on purpose: the section is a snapshot of one measurement session at a named revision, and marking the item done without re-measuring would put an unmeasured claim into a document whose value is that every number in it was measured through the committed driver. Resolving it honestly needs a benchmark re-run, which is outside a documentation pass.

🔧 Fix: Deduplicate architecture report onto .tex, correct columnar sections
1 info still open:
- ℹ️ `docs/SIRIX_ARCHITECTURE_REPORT.tex:5025` - Scope judgment call, already applied. §8.4&#39;s fail-closed list claimed non-double-exact numerics (BigDecimal/BigInteger/float) throw and defer to the interpreter. That is wrong against the code — classifyNumberForPredicate returns FK_DECIMAL for BigDecimal and BigInteger, and only xs:float throws — but FK_DECIMAL predates this branch (15 occurrences at base commit 5ce555630), so the staleness is not strictly this change&#39;s doing. I corrected it anyway rather than leave it: the §8.2 text I was instructed to write states there is an exact-decimal column, and an uncorrected §8.4 three paragraphs later would have read as a direct contradiction introduced by my own edit. Flagging in case you want the correction reverted to keep this commit strictly limited to what the branch made stale.

</details>

<details>
<summary>⚠️ **Lint** - 1 info</summary>

- ℹ️ `build.gradle:163` - Spotless (eclipse formatter, eclipse/sirix-formatter.xml) reports violations in 241 files repo-wide, and the build deliberately sets `enforceCheck false` so formatting is not gated. I checked my own changed file in isolation via `-PspotlessIdeHook`: Spotless wants ~14,700 line changes in SirixVectorizedExecutor.java alone (a 13.6k-line file), i.e. it reflows every javadoc paragraph — so the file was already wholesale non-conforming before this pass, and my 6-line comment edit introduces nothing new. I did not run spotlessApply: it would reformat the entire codebase, which is the broad mechanical migration this pass is scoped to avoid, and it would bury this change&#39;s diff. Unresolved by design; a repo-wide `spotlessApply` (or a decision to keep formatting unenforced) is a separate piece of work.

</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.

</details>


JohannesLichtenberger:
Pushed `d3af001`, which resolves the "**`docs/COMPARISON_POSTGRES_BULK.md` is stale**" item from the PR description. The doc is re-measured on a freshly rebuilt 2.12 GB corpus (3,482,208 records, all three arms cross-checked identical). Two of the corrections changed conclusions rather than digits.

### The memory accounting was wrong

The setup table said `-Xmx8g` against `shared_buffers=1GB`. That compares a JVM heap to a buffer pool, and `-Xmx` does not describe SirixDB's footprint — its page cache is **off-heap**, sized from `DatabaseConfiguration`'s segment-allocation budget (default 16 GB) and split by `initializeGlobalBufferManager`. Read back from a live process:

| SirixDB cache (off-heap) | bytes |
|---|---:|
| `recordPageCache` | 8,589,934,592 (8.00 GB) |
| `recordPageFragmentCache` | 3,221,225,472 (3.00 GB) |
| metadata `pageCache` | 1,073,741,824 (1.00 GB) |

So the published run gave SirixDB an 8 GB record cache — which holds the entire 2.12 GB corpus — and PostgreSQL 1 GB. It compared a *cache-resident* SirixDB against a *buffer-pressured* PostgreSQL. Every query number is now measured twice at matched sizes: regime A ~1 GB each, regime B 8 GB each.

### The buffer-pressured regime was never achieved

A and B land within noise of each other, because the OS page cache defeats the experiment: 2.12 GB sits in a 16 GB machine's page cache whatever `shared_buffers` says, so shrinking it moves reads from PostgreSQL's pool to the kernel's, not to the device. Cold proves it — `countAll` is 279 ms warm and **2,127 ms** straight after `drop_caches`, a 7.6× gap the 1 GB-vs-8 GB setting doesn't come close to reproducing. The regime table now reads **NOT ACHIEVED**; reaching it needs a corpus larger than RAM.

### Warm scans reverse

| query | previous revision | now (regime A) | PG `jsonb` | PG normalized |
|---|---:|---:|---:|---:|
| `countAll` | 202 ms | **93 ms** | 279 ms | 255 ms |
| `filterCountYear` | 19,321 ms | **237 ms** | 388 ms | 264 ms |
| `sumYear` | 18,768 ms | **223 ms** | 399 ms | 260 ms |
| `titleLookup` | 19,460 ms | 283 ms | 354 ms | **250 ms** |

The previous revision concluded PostgreSQL won these by 45–78×. SirixDB now beats the like-for-like `jsonb` arm on all four in both regimes, and beats or matches the normalized arm on three of four.

### Cold does not

| query | previous | now | PG `jsonb` | ratio |
|---|---:|---:|---:|---|
| `countAll` | 40,566 ms | 54,321 ms | 4,974 ms | PG **11×** |
| `filterCountYear` | 84,606 ms | 57,824 ms | 1,969 ms | PG **29×** |
| `sumYear` | 76,274 ms | 46,660 ms | 3,732 ms | PG **13×** |
| `titleLookup` | 82,408 ms | 57,232 ms | 2,860 ms | PG **20×** |

Consistent with what the warm fixes actually were — they stop the engine discarding a warm cache, which does nothing for a cache that starts empty. A warm scan is now ~200–600× faster than the same scan cold, and **cold is the largest remaining gap in the comparison**.

### What was and was not re-run

Storage was re-verified and reproduces: `movies_rel` came back **byte-identical** at 2,150,776,832, `movies_jsonb` within 0.001%, and a default SirixDB store slightly smaller than before. **Ingest (§3) and the projection-vs-B-tree comparison (§4.2) were not re-measured** and now say so inline, rather than letting carried-over numbers read as current — §4.2 in particular is likely to have moved, since the same engine changes shifted §4.1 by ~70×.

The two remaining caveats in the PR description still stand as written: `formal-verification.md` was not reviewed, and the `Names.removeName` probe-chain limitation is documented rather than fixed.

---

PR #1170, opened 2026-08-15
Port ClickBench to SirixDB, and fix the three wrong answers it found

JohannesLichtenberger:
Ports [ClickBench](https://github.com/ClickHouse/ClickBench) — 43 analytical SQL queries over one 105-column web-analytics table — to SirixDB, and fixes the three wrong-answer defects the port found on the way.

## The port

`docs/CLICKBENCH.md` is the writeup. In short:

```bash
./gradlew :sirix-query:clickBenchLoad -Pclickbench.args="/tmp/cb-1m generate:1000000"
./gradlew :sirix-query:clickBench     -Pclickbench.args="/tmp/cb-1m --tries 3 --json /tmp/cb.json"
bundles/sirix-query/bench/clickbench/run-differential.sh 200000   # the correctness gate
```

The real dataset works too (`prepare-data.sh hits.parquet out.json [rows]`); the synthetic generator exists so the port runs without the 14 GB download, and it plants the literals Q19/Q40/Q41 filter on so those queries are not trivially empty.

Three translations are forced by the engine rather than chosen:

| SQL | JSONiq | why |
|---|---|---|
| `LIMIT n OFFSET k` | `fn:subsequence(expr, k+1, n)` | `[...]` is an array index in JSONiq, not a positional predicate — and this is the shape the top-k pushdown keys on |
| `HAVING` | a `where` after `group by` | XQuery 3.0 free clause ordering |
| `DATE_TRUNC('minute', t)` | `substring($t, 1, 16)` | timestamps are ISO-8601 strings, which order lexicographically, so `ORDER BY EventTime` and the `EventDate` ranges stay string comparisons |

The loader re-reads the first record and **fails the load** if a 64-bit id arrived as a string. ClickHouse's `JSONEachRow` quotes 64-bit integers by default, so the official `hits.json.gz` may carry `"UserID":"435090932899640449"` — that shreds as a string node, and then exactly three queries quietly return nothing while the other forty look plausible.

## The three defects

Each served a plausible number rather than crashing; all were found by differential testing.

1. **`sum`/`avg` over 64-bit integers wrapped at 2^64.** `AVG(UserID)` over 200k records answered `1.67921e13` where the true average is `5.71681e17` — the reported sum was exactly the true sum modulo 2^64. The kernels now detect overflow (exactly in the scalar lanes, by a magnitude bound in the SIMD page kernel) and redo the fold exactly; an overflowed `long[]` never reaches the `(source, field)`-keyed cache. The same wrap existed a second time in `PathStats`, which serves `sum`/`avg` directly whenever a path summary is on.
2. **`min`/`max` over a string column threw instead of answering.** `requireSupported` turns a decline into an error, so there was nothing to fall back to; `MIN(EventDate)` was simply unrunnable in the default configuration. String extrema are well defined, so the kernel answers them now. `sum`/`avg` over strings stay loud.
3. **An aggregate over a nested `for` on a grouped variable folded the whole input.** `sum(for $x in $g return $x.f)` in a group-by return was answered with the global fold, once per group; `count($g)` builds no inner pipeline and stayed correct, which is what made it hard to notice. `RegroupedSourceGuardStage` withdraws a claim whose scan source is a variable an enclosing `group by` has rebound. The root cause is upstream (see the companion brackit PR); fixing it there makes this stage redundant rather than wrong.

## Verification

* Three-way differential over byte-identical records — SirixDB fast paths, SirixDB interpreter, DuckDB — **0 mismatches on both legs** at 200k rows. Roughly a quarter of the queries are `ORDER BY <agg> DESC LIMIT 10` over data with many equal counts, so tie membership at the cut-off is engine-defined (DuckDB itself varies across runs); `compare-results.py` accepts a difference only when every differing row is tied at the window boundary, and marks a window lying entirely inside one tied group as *unverifiable* rather than counting it as checked.
* `TypedGroupByDifferentialTest` (96 assertions, every query through both pipelines) and `VectorizedAggregateExactnessTest`, whose expected values come from `BigInteger` rather than from the engine.
* `:sirix-query:test` (1155 tests) and `:sirix-core:test` green.
* No measurable cost: the ClickBench suite's hot total moved 76.2 → 75.6 s across the fixes.

## Measured

1M synthetic rows, 20 cores, no projection index: load 62.7 s for 1.36 GB, cold total 84.6 s, hot total 75.6 s over the 43 queries. That is 1M, **not** ClickBench's 100M, which extrapolates to ~1.7 h and ~136 GB and has not been run. The resource is loaded in the fast-ingest configuration — no path summary, no projection index — so these are the untuned row-path numbers; applying the projection index to the columns these queries touch is the obvious next step.

## Note for review

`spotlessJavaCheck` could not be run in this environment (the eclipse-formatter artifacts are not in the offline Gradle cache). The profile's mechanical rules were checked by hand on every touched file; please run `./gradlew :sirix-query:spotlessApply :sirix-core:spotlessApply` if CI's ratchet complains.

https://claude.ai/code/session_01464v97RZ1FGrnsAuww83P3


JohannesLichtenberger:
Closing as superseded: the ClickBench changes are already represented in origin/main (87 patch-identical commits plus the amended main equivalents for the two ingestion commits). No unique code from this PR remains to merge.

Issue #45, opened 2018-12-26
JSON node layer


JohannesLichtenberger:
Each node has a unique 64bit identifyer (long).

- Object node, consisting of record-nodes (simple name/value)-pairs. The value itself must be a another node, referenced by a long.
- Array-node, consitsing of arrayItem-nodes, consisting of any type of nodes (long-values). We can simply store pointers to the right sibling for each arrayItem-node.
- Null-node.
- Boolean-node.
- String-node.
- Number-node, maybe the same as string but with a bit set, if it's a number or a string.

This simple design allows fine-granular copy-on-write operations.

PR #1184, opened 2026-08-26
feat(sirix-core): maintain every index family in the parallel bulk import's single pass

JohannesLichtenberger:
## Intent

One landing for the bulk-import index campaign. The parallel JSON importer now maintains every index family it can feed in the load's single pass: projection rows are extracted in the build workers from the primitives they already hold and replayed in document order through the unchanged leaf/dictionary machinery (byte-identical to the post-pass by a 4020-slot census; 1M one-pass fell 497 s to 23.8 s, 1.9x bare import and 2.8x faster than bare-then-post-pass), and PATH/CAS/NAME tuples ride the same worker hooks into the families' bulk loaders (sequential-oracle-identical across 20 probes; +10.2% import cost for four real defs). Order labels mint through a lookup-free in-order lane that also cures the post-pass's first-label full-corpus mintRun walk; intermediate TIL epoch flushes bound its retention (~330k pinned pages before leaf 1 at 100M previously); the HOT bulk splice packs leaves canonically (a pre-existing leaf-packing bug produced 7x fragmentation in every PATH/CAS/NAME bulk build; fixed with the invariant battery green). From the review branch: executor retirement moves the pool join off the compile path onto a shared reaper (closing a verified ~1G-per-retired-executor native retention: the fresh-executor A/B now runs flat where it previously climbed to a kernel OOM kill) with the warm-up queue drained rather than interrupted, and the slab allocator reserves an oversized headroom its ratchet can never consume. Fat string columns now serve through windowed 128-leaf loads under an eagerMaterializeBytes budget instead of whole-column eager materialization: at 100M, q5 completes in 48.1 s and q20 in 17.5 s inside an 8G-heap envelope where every prior attempt OOMed, and an over-budget column fill declines through the store's established non-memoizing door (deliberately not marked as corruption). The #45/#50/#52/#53/#55 set lands adapted to current semantics: stale tombstones carry machine-readable reasons and the one silent load-degradation (dictionary budget breach) prints unconditionally to stderr; the HOT fragment-merge/carry-forward/in-place-split paths gain flag-gated counters whose sharpest member is a permanent sentinel on the complete-dump break; three of its pieces were dropped as structurally superseded (chunked-arena growth clamp, interval-sweep locate, refusal-site counters), each with the stronger in-tree equivalent named in the commit. The streaming-dictionary follow-on (resident probe front, writer dict-record memo, Unsafe-free XXH3) is pre-existing committed history in this lineage, not part of this delta. Two latent defects found and fixed on the way: builders' resolved-path-class caches going stale under a growing summary, and the cached valid-time flag bypassing catalogued-but-unbound defs. Valid-time maintenance still refuses, now def-based. Deliberate choices a reviewer should not flag: no defensive guard on the post-pass intermediate flushes (an earlier guard was deleted after a 40k-record forced-rotation experiment proved flushed pages of the open transaction read back exactly); whole-build splice-once was rejected by memory arithmetic in favor of an 8M-entry/512MB accumulate-with-cap; the rider deliberately does NOT use the bulk slot loader yet (needs a pre-commit finalize hook, documented as v2); intermediate commits of the series are review units — the tip is the fully-gated tree (10,570 core + 1,466 query tests green, plus the 100M acceptance runs).

## What Changed

- **One-pass index maintenance in the parallel bulk JSON importer.** New `ParallelBulkJsonImporter`/`BulkJsonTreeAssembler` loaders (feeder scan, worker page builders, ordered adoption, NDJSON adapter) now extract projection rows *and* PATH/CAS/NAME tuples in the build workers from the primitives they already hold; the coordinator drains them in document order into the families' ordinary bulk loaders, with one flush per family before the caller's commit. A catalogued projection with no armed load-time build is refused rather than silently left unmaintained; valid-time interval maintenance still refuses, now def-based. The post-pass projection build gains a lookup-free in-order label lane and intermediate TIL epoch flushes to bound retention, and `HOTBulkBuilder`'s splice now packs leaves canonically — a pre-existing fit-bound bug had been fragmenting every PATH/CAS/NAME bulk build.
- **Wide projection string columns serve through bounded windowed leaf loads.** Columns whose worst-case resident size exceeds `-Dsirix.projection.eagerMaterializeBytes` are read from 128-leaf windows via `ProjectionWindowedRowGroupPayloads`/`ProjectionColumnStore` instead of whole-column eager materialization; an over-budget fill declines through the store's non-memoizing door as a routing decision (the windowed whole-leaf route or the record path answers), not a corruption signal, so the index stays valid.
- **Reliability and diagnostics.** Vectorized executor retirement moves the pool join off the compile path onto a shared daemon reaper with the warm-up queue drained rather than interrupted; `FrameSlotAllocator` reserves an oversized-allocation headroom that permanent slab commitment can never consume; stale projection tombstones carry machine-readable reasons on previously reserved flag bits (`PIXM` wire format unchanged) and a dictionary-budget breach prints `[proj] PROJECTION ABANDONED` with the breaching quantity unconditionally to stderr; HOT fragment-merge/carry-forward/in-place-split paths gain flag-gated counters plus a permanent sentinel on the complete-dump break. Two latent defects fixed: builders' resolved-path-class caches going stale under a growing path summary, and the cached valid-time flag bypassing catalogued-but-unbound definitions. Adds `docs/BULK_IMPORT.md` and `docs/HOT_BULK_BUILD.md`, plus equivalence/oracle test suites for bulk assembly, primitive-index and projection parity, windowed serving, and HOT dump-merge and tombstone eviction.

## Risk Assessment

⚠️ Medium: The delta is very large and its memory-budget/windowed-serving machinery has been rewritten across eleven rounds, but the current state has no substantiated reachable correctness defect — every fill door prices and charges through one shared incremental helper, the windowed view carries no session-derived state, and every route that can receive a budget decline now treats it as routing; the two residual findings are an uncounted decline plus misattributed javadoc, both safe as follow-ups.

## Testing

I exercised the dictionary-budget abandon path end-to-end through real load-time builds and confirmed the operator notice now prints coherent arithmetic. The decisive evidence is a same-corpus, same-calibrated-budget A/B: with the four production files reverted to the pre-fix commit the notice reads "retained 1114240 B ... past its 39850930 B budget" and the loud arm fails, while the commit under test prints "needed 39854408 B (flush-peak) ... past its 39850930 B budget (1114240 B retained)" and passes — the announced quantity genuinely exceeds the announced budget, the tripping term is named, and retention is demoted to context. I also confirmed the exception message behind the notice now quotes a budget worth raising to. Because the notice test reaches only the flush-peak refusal family, I extended GlobalValueDictionaryWriterBudgetTest to assert the byte-decline invariant directly on the exception and added a case covering the structural-ceiling family, which weighs no bytes and must not claim a breach — previously unpinned, and the one way this fix could silently regress. Twenty focused tests pass across the notice, writer-budget, stale-reason and generation-rotation suites; the worktree is clean apart from the intentional test addition. No screenshot artifact applies: the changed surface is a stderr line emitted by a database loading process, with no UI, HTML or rendered output, so the captured terminal transcript is the actual end-user experience.

<details>
<summary>Evidence: Operator stderr notice: before/after the breaching-quantity fix (same corpus, same budget)</summary>

<code>BEFORE (pre-fix production code):&#10;[proj] PROJECTION ABANDONED during the load: index 0, column 0 retained 1114240 B over 10227 distinct values, past its 39850930 B budget. The load completes; the projection is STALE and every query will take the generic pipeline until jn:create-projection-index rebuilds it.&#10;-&gt; 1,114,240 B declared &#34;past&#34; a 39,850,930 B budget: a number 36x larger. Test FAILS.&#10;&#10;AFTER (12dc10bc8):&#10;[proj] PROJECTION ABANDONED during the load: index 0, column 0 needed 39854408 B (flush-peak) over 10227 distinct values, past its 39850930 B budget (1114240 B retained). The load completes; the projection is STALE and every query will take the generic pipeline until jn:create-projection-index rebuilds it.&#10;-&gt; 39,854,408 B &gt; 39,850,930 B. Term named. Retention kept as context. Test PASSES.</code>

```text
Operator-facing evidence: the dictionary-budget abandon notice now prints the quantity
that actually breached, so its own arithmetic explains the breach it announces.

Surface: stderr of the loading process (System.err, deliberately not a log appender, so the
shipped logback root-at-ERROR configuration cannot swallow it).

Both transcripts below come from THE SAME test, THE SAME 30,000-record corpus and THE SAME
budget, which the test calibrates from the real writer rather than hard-coding.  Only the
production code differs, so the numbers are directly comparable.

-----------------------------------------------------------------------------------------
BEFORE  (production files reverted to c17d3b378, the commit before the fix)
-----------------------------------------------------------------------------------------
$ ./gradlew :sirix-core:test --rerun \
      --tests 'io.sirix.index.projection.ProjectionDictionaryBudgetAbandonNoticeTest'

[proj] PROJECTION ABANDONED during the load: index 0, column 0 retained 1114240 B over 10227
distinct values, past its 39850930 B budget. The load completes; the projection is STALE and
every query will take the generic pipeline until jn:create-projection-index rebuilds it.

    ^ 1,114,240 B is declared to be "past" a 39,850,930 B budget -- a number 36x LARGER.
      The operator is told a guard fired, shown arithmetic that says it should not have,
      and given no figure to raise the budget to.

ProjectionDictionaryBudgetAbandonNoticeTest >
    aDictionaryBudgetBreachAnnouncesTheAbandonedProjectionEvenWithLoggingOff() FAILED
    org.opentest4j.AssertionFailedError: the notice no longer states index, column, the
    breaching term and the budget: [proj] PROJECTION ABANDONED during the load: ...
        at ProjectionDictionaryBudgetAbandonNoticeTest.java:163
2 tests completed, 1 failed
BUILD FAILED

-----------------------------------------------------------------------------------------
AFTER  (12dc10bc8, the commit under test)
-----------------------------------------------------------------------------------------
$ ./gradlew :sirix-core:test --rerun \
      --tests 'io.sirix.index.projection.ProjectionDictionaryBudgetAbandonNoticeTest' \
      --tests 'io.sirix.index.projection.GlobalValueDictionaryWriterBudgetTest' \
      --tests 'io.sirix.index.projection.ProjectionStaleReasonTest'

[proj] PROJECTION ABANDONED during the load: index 0, column 0 needed 39854408 B (flush-peak)
over 10227 distinct values, past its 39850930 B budget (1114240 B retained). The load
completes; the projection is STALE and every query will take the generic pipeline until
jn:create-projection-index rebuilds it.

    ^ 39,854,408 B > 39,850,930 B.  The announced quantity really does exceed the announced
      budget; the term that tripped the predicate is NAMED (flush-peak, i.e. the projected
      cost of flushing, not current occupancy); and retention is kept as context rather than
      being passed off as the thing that breached.

BUILD SUCCESSFUL
  ProjectionDictionaryBudgetAbandonNoticeTest    tests=2  failures=0  errors=0
  GlobalValueDictionaryWriterBudgetTest          tests=13 failures=0  errors=0
  ProjectionStaleReasonTest                      tests=5  failures=0  errors=0

Note the tripping term: retention was only 2.8% of the number that actually failed the
comparison, which is precisely why quoting retention read as a misfiring guard.
```
</details>
<details>
<summary>Evidence: Exception messages per decline family (real writer, no fault injection)</summary>

<code>=== BYTE-BUDGET BREACH ===&#10;breachingTerm = flush-peak ; breachingBytes = 510680 ; budgetBytes = 509824 ; retainedBytes = 118400 ; breaching &gt; budget ? true&#10;message: Resource-wide value dictionary for column 3 reached its configured aggregate budget of 509824 B; its flush-peak came to 510680 B, past the 509824 B budget, with 118400 B already retained across 103 distinct values. ... Raise the configured byte budget to at least 510680 B, use per-leaf dictionaries, or split the ingest into bounded append generations.&#10;&#10;=== STRUCTURAL DECLINE (nothing weighed in bytes) ===&#10;breachingTerm = null ; breachingBytes = 51072 (== retained)&#10;message: Resource-wide value dictionary for column -1 declined an unsafe allocation before mutation: value length 262145 exceeds the safe V0 limit of 262144 bytes; retained 51072 B across 0 distinct values. ... raising the byte budget cannot override a structural allocation ceiling.</code>

```text
Operator-facing evidence: the exception message behind the notice, per decline family.
Produced by driving the real GlobalValueDictionaryWriter to each decline through its own
API (no fault injection) and printing what the operator would read in the log.

=== BYTE-BUDGET BREACH (the family the abandon notice reports) ===
budgetBytes   = 509824
breachingTerm = flush-peak
breachingBytes= 510680
retainedBytes = 118400
breaching > budget ? true
message: Resource-wide value dictionary for column 3 reached its configured aggregate budget of 509824 B; its flush-peak came to 510680 B, past the 509824 B budget, with 118400 B already retained across 103 distinct values. The projection is abandoned for this load (readers fall back to the generic pipeline); the load itself completes. Raise the configured byte budget to at least 510680 B, use per-leaf dictionaries, or split the ingest into bounded append generations.

=== STRUCTURAL DECLINE (nothing was weighed in bytes) ===
budgetBytes   = 9223372036854775807
breachingTerm = null
breachingBytes= 51072  (== retained 51072)
message: Resource-wide value dictionary for column -1 declined an unsafe allocation before mutation: value length 262145 exceeds the safe V0 limit of 262144 bytes; retained 51072 B across 0 distinct values. The projection is abandoned for this load (readers fall back to the generic pipeline); the load itself completes. Use per-leaf dictionaries or split the ingest into bounded append generations; raising the byte budget cannot override a structural allocation ceiling.
```
</details>
<details>
<summary>Evidence: Focused test results (JUnit XML summary)</summary>

```text
GlobalValueDictionaryWriterBudgetTest.xml: tests=13 skipped=0 failures=0 errors=0
ProjectionDictionaryBudgetAbandonNoticeTest.xml: tests=2 skipped=0 failures=0 errors=0
ProjectionStaleReasonTest.xml: tests=5 skipped=0 failures=0 errors=0
new case present: testcase name="A structural ceiling declines without claiming any quantity exceeded the byte budget"
ProjectionGlobalDictionaryGenerationRotationTest: BUILD SUCCESSFUL
#67 WITNESS: distinct=36864 (ceiling 16384), records=147456 | ONE-PASS: stale=false columnKind=5 rowGroups=144 | POST-PASS: THREW IllegalStateException: Forced global value dictionary for column 0 cannot continue safely: append entry 16,385 exceeds the safe per-append limit of 16,384
```
</details>
- Outcome: 🔧 1 issue found → auto-fixed (3) ✅ across 4 runs (1h1m6s)

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

_... (7 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>
<summary>⚠️ **Review** - 2 infos</summary>

🔧 Fix: Per-caller window views, cumulative fill budget, honest declines
4 issues (1 warning, 3 infos) still open:
- ⚠️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionColumnStore.java:1842` - This round gave every column-bytes door a budget check and a ledger charge — `column` (:790, :806), `columnMasked` (:1287), `columnBytes` (:1571, :1582), `columnDistinctIdentity` (:1085, :1099) — and the DATA weigher&#39;s new comment (ProjectionIndexCatalog.java:147-152) rests the flat 1.25x charge on that: &#34;the store&#39;s CUMULATIVE retained-fill budget caps that sum at eagerMaterializeBytes, which is what makes this flat charge an upper bound rather than a hope&#34;. `recordKeys(ColumnSegmentFetcher)` is the one fill door that got neither. It calls `fetchSegmentChain(-1, keysColumnSegmentId(), SEG_KIND_KEYS, false, fetcher)` with NO keep mask — every leaf&#39;s KEYS segment at once — and publishes the decoded `long[][]` into `recordKeySlices` (:290), retained for the store&#39;s whole cache lifetime, charged nothing. Concrete path on the workload the intent advertises: a 100M-row windowed handle, `for $r in $doc[] where $r.SearchPhrase eq &#34;...&#34; return $r` reaches `predicateScanRecordKeys`, whose gate `predsSliceable` is kind-only (SirixVectorizedExecutor.java:2788) and admits it; `ProjectionColumnScan.matchingRecordKeys` (:439/:554/:769) calls `store.recordKeys(fetcher)`, which fetches ~800 MB of raw KEYS bytes and retains a decoded `long[100M]` (~800 MB) beside the ~2 GB the ledger permits — i.e. ~1.6 GB of transient plus ~800 MB of permanent residency that the weigher&#39;s 2.5 GB charge does not cover, on the 8 G heap the change advertises as OOM-free, with four such handles admissible concurrently. `bloomBytes` (:217) and `stringExtrema` (:300) are smaller instances of the same omission. Fix at the boundary the round already established: price the KEYS fill through `checkFillBudget` and charge `retainedFillBytes` on publish, exactly as the four sibling doors do — an over-budget KEYS chain then declines through `FillBudgetExceededException`, which the predicate-scan callers already answer by falling back.
- ℹ️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionIndexCatalog.java:868` - `windowedResidentWeightBytes` clamps the EAGER branch to `Math.max(1L, CACHE_BYTES &gt;&gt; 1)` (:873) to stop a handle heavier than `maximumWeight` from self-evicting on insert — but that branch is only reached when `projectedWeightBytes &lt;= eagerMaterializeBytes`, and `eagerMaterializeBytes` defaults to `min(CACHE_BYTES / 2, maxMemory / 4)` (:755-756), so the value is already at or below the clamp: the clamp is dead by construction. The WINDOWED branch (:868), which returns `eagerMaterializeBytes * 1.25`, carries no clamp at all — and it is the one that can exceed the cache&#39;s own `maximumWeight` (`max(1, CACHE_BYTES &gt;&gt; 10)` units, :141-143) when `-Dsirix.projection.eagerMaterializeBytes` is set above `cacheBytes / 2`: e.g. `-Dsirix.projection.eagerMaterializeBytes=17179869184` with the default 8 GiB `cacheBytes` charges 20 GiB (weight 20,971,521 &gt; 8,388,608), reproducing exactly the insert-time self-eviction and per-lookup re-decode this method was changed to prevent. Clamp the returned value once, after the branch, rather than inside the branch that cannot exceed it.
- ℹ️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:13391` - The two new budget-decline handlers diverge on their terminal case. `groupByAggregate`&#39;s returns `declineGroupAgg(&#34;column fill over budget on the whole-leaf route too&#34;)` (:13129), which ticks GROUP_AGG_DECLINED and logs a reason. `constGroupAggregate`&#39;s returns a bare `null` (:13391), so a const-group query that declines on BOTH passes and falls to the generic navigational pipeline is invisible in every counter — neither GROUP_AGG_DECLINED nor GROUP_AGG_FAILED moves. That is the same observability hole the round-5 finding raised about the first version of these handlers, left standing on this one arm. Route it through the arm&#39;s own decline helper (or tick the decline counter) so the double-decline is countable.
- ℹ️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionWindowedRowGroupPayloads.java:61` - `ReaderSource`&#39;s javadoc still describes the design this round replaced: &#34;each CONSULT rebinds the source to the calling reader&#39;s own live session ({@code bindReaderSource}), and a window fetch opens its transaction through whatever source is bound at that moment&#34;. `bindReaderSource` no longer exists, and &#34;whatever source is bound at that moment&#34; is precisely the last-writer-wins semantics the round removed — the class javadoc twenty lines above (:41-47) now states the opposite and correct contract. A reader who reaches this block first is told the shared cache carries a mutable binding. Rewrite it to describe `boundTo`/`BoundView` (source lives on the per-caller view, passed into `payload(index, source)` per access).

🔧 Fix: Price record-key fills, clamp weigher once, count const declines
2 issues (1 warning, 1 info) still open:
- ⚠️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionColumnStore.java:876` - The whole fill-budget system prices RAW segment bytes, but the residency it exists to bound is raw + the decoded arrays — and the change&#39;s own code prices those decoded arrays everywhere else. `projectedColumnFillBytes(col)` (:876-903) sums only `RowGroupDescriptor.entryByteLen` for BODY and DICT; its own javadoc (:871-874) says so (&#34;An UNDERestimate of the resident total (decoded id/value arrays come on top)&#34;). `ProjectionIndexCatalog.residentWeightOf` (:661-682), which computes `projectedWeightBytes` for the SAME columns, explicitly adds `((long) rows &lt;&lt; 3) + presenceBytes` for every `isLongLaneKind` column — the 8 B/row decoded lane. And this round&#39;s new `projectedRecordKeysFillBytes` (:692-699) counts `8L * rowCount` per leaf with a comment stating exactly why (&#34;Unlike the column doors this counts the decoded form explicitly, because that is the part that stays&#34;). The column doors are the exception, and they are the ones the budget rests on. Concrete path on the advertised 8 G heap (`columnFillBudgetBytes` = `eagerMaterializeBytes` = min(cacheBytes/2, maxMemory/4) = 2 GB): a 100M-row windowed handle with bit-packed NUMERIC_LONG columns, each BODY ~100 MB packed. `columnFillable`/`columnFillWithinBudget` (:762-768) admit ~20 such columns before `retainedFillBytes` reaches 2 GB; `column()` retains for each of them BOTH the raw `byte[][]` (charged, via `columnBytes` at :1610-1622) AND the decoded `ColumnSlice[]` whose `numericValues()` is a `long[rows]` — ~800 MB per column, ~16 GB total, none of it priced. Nothing declines, and the heap the intent advertises as OOM-free is exhausted. The same understatement propagates to the DATA weigher: `ProjectionIndexCatalog.java:147-154` charges a windowed handle a flat 1.25x `eagerMaterializeBytes` and rests that on the ledger (&#34;the store&#39;s CUMULATIVE retained-fill budget caps that sum at eagerMaterializeBytes, which is what makes this flat charge an upper bound rather than a hope&#34;) — a claim that is false while the ledger counts packed bytes and the heap holds decoded ones, so up to four such handles are admitted concurrently. Fix at the pricing boundary the rest of the file already uses: have `projectedColumnFillBytes` (and `projectedMaskedFillBytes`, `projectedColumnIdentityFillBytes`) add the decoded lane the same way `residentWeightOf` does — `(rows &lt;&lt; 3) + presenceBytes` per long-lane column over the leaves the fill touches — so one budget figure means the same thing on both sides of the cache.
- ℹ️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:13159` - Two fix rounds re-created the exact defect round 1 reported and the author fixed in `Names`/`PathSummaryWriter`/`StorageEngineWriter`: a new member inserted between an existing doc block and the member it documented. (1) `SirixVectorizedExecutor.java:13159` — the block `/** Constant-key group-bys served in one scalar pass (test oracle, same doctrine as the rest). */` now precedes `declineConstGroupAgg`&#39;s own javadoc and method, leaving `CONST_GROUP_AGG_SERVED` (:13176) undocumented and stacking two doc blocks on one method. (2) `ProjectionIndexCatalog.java:848-855` — the block documenting `servesWindowedPayloads` (&#34;Whether a handle of this projected weight is served by the WINDOWED payload view...&#34;, with its own `@param`/`@return`) now precedes `windowedResidentWeightBytes`&#39;s javadoc and method, leaving `servesWindowedPayloads` (:879) undocumented. Generated javadoc attributes each stale block to the wrong member, and the `@param projectedWeightBytes` in the catalog block duplicates the one below it. Move each new method below the member its preceding block belongs to.

🔧 Fix: Price fills by decoded residency through one shared function
2 issues (1 warning, 1 info) still open:
- ⚠️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionColumnStore.java:898` - The new shared pricing function returns 0 for the exact column kind the change exists to serve. `decodedColumnResidentBytes` (:886-899) prices the 8 B/row lane + presence for `isLongLaneKind` and two presence words for BOOLEAN, and falls through to `return 0` for `COLUMN_KIND_STRING_DICT` (2) and `COLUMN_KIND_STRING_SET` (4). But a STRING_DICT fill retains, per leaf: `int[] ids` from `decodePackedIds` — 4 B/row (ProjectionIndexRowGroupCodec.java:697-704), `long[] presence` (presenceBytes), `int[] dictOffsets` — 4 B per dict entry, and for `DICT_MODE_FSST`/`DICT_MODE_FSST_ROW_COUNTS` a freshly DECOMPRESSED dictionary `Arrays.copyOf(scratch, pos)` (ProjectionIndexColumnSegmentCodec.java:2432, reached from decodeStringSlice at :1660-1662) that is several times the compressed DICT segment bytes the budget counts. The file&#39;s own diagnostic `sliceRetainedBytes` (:1529-1556) enumerates precisely these terms, so the formula is already known here. Concrete path on the 8 G heap the intent advertises (`columnFillBudgetBytes` = `eagerMaterializeBytes` = min(cacheBytes/2, maxMemory/4) = 2 GB): a 100M-row `SearchPhrase`/`URL` STRING_DICT column whose BODY+compressed-DICT projection prices at ~1.9 GB passes `columnFillWithinBudget` (:766-769) and `column()` retains ~400 MB of id lanes plus an FSST-expanded dictionary on top of it — and because round 7 wired the DATA weigher to this same function (ProjectionIndexCatalog.java:667-671), the cache ADMITS handles on the same understated figure while its comment (:147-154) calls the flat charge &#34;an upper bound rather than a hope&#34;. This is the identical defect class the round fixed for long lanes, left standing on the kind q5/q20 exercise. Secondary evidence that the string side was not considered: the loop this round added to `projectedColumnIdentityFillBytes` (:1087-1089) runs only for a column that is STRING_DICT by construction (early return at :1044-1048), so it contributes exactly 0 — dead arithmetic — while the identity fill&#39;s real decoded residency (4 B/row ids + 8 B/entry `dictHashes`) goes uncounted. Fix inside the shared function: add 4 B/row + presence for STRING_DICT/STRING_SET, 4 B per dict entry for the offsets, and a conservative expansion multiple on the DICT segment bytes for the FSST modes.
- ℹ️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionColumnStore.java:838` - The cumulative retained-fill ledger counts every column&#39;s BODY chain twice. `column(col, fetcher)` prices `projectedColumnFillBytes(col)` — BODY + DICT + decoded (:823) — and charges it at :838; the fill it performs starts with `fillColumn` (:1425) → `columnBytes(col, fetcher)` (:1431), which independently prices `projectedColumnBodyFillBytes(col)` (:701-717, the BODY chain alone) and charges it at :1660 into the same monotonic `retainedFillBytes` AtomicLong (:991, never decremented). One retained copy of the BODY arrays, two charges. `columnDistinctIdentity` has the same shape: it charges `projectedColumnIdentityFillBytes` (:1156/:1170), whose walk includes the BODY chain, while `fillIdentityColumn` calls `columnBytes` at :1213. Effect is over-conservative rather than wrong: after one fill the ledger reports more residency than exists, so `columnFillWithinBudget` (:766-769), `columnFillable` (:788) and `columnIdentityFillable` (:1105-1108) decline the next viable fill earlier than the budget intends and the planner steers a servable query to the slower whole-leaf route — never a wrong answer, but the campaign&#39;s whole point is route quality, and `retainedFillBytes()` is also the number the new tests read. Charge the delta instead: have `column()`/`columnDistinctIdentity` add only what `columnBytes` did not already charge (e.g. `projected - projectedColumnBodyFillBytes(col)`).

🔧 Fix: Price string fills by decoded residency, charge ledger once
2 issues (1 warning, 1 info) still open:
- ⚠️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionColumnStore.java:824` - Round 9 made the ledger CHARGE each retained byte once (`chargeRetained`, :1055, subtracting `projectedColumnBodyFillBytes` when `columnBytes[col]` is already published) but left the budget CHECK and the planner predicates pricing the GROSS projection against that same ledger, so a body chain already counted in `retained` is counted a second time in `projected`. Concrete reachable path: `ProjectionColumnSegmentFoldScan` fills raw BODY bytes directly through `store.columnBytes(col, fetcher)` for long-lane columns (:196, :318, :763, :916) — the fused fold route for a plain `sum($r.age)` — which publishes `columnBytes[col]` and charges `projectedColumnBodyFillBytes(col)` at ProjectionColumnStore.java:1724. A later query on the SAME cached handle takes the sliced group route on that column and reaches `column(col, fetcher)`, whose `checkFillBudget(col, projected, &#34;slice fill&#34;)` at :824 evaluates `retained + projectedColumnFillBytes(col) &gt; columnFillBudgetBytes`. Numbers on the 8 G heap the intent advertises (`columnFillBudgetBytes` = `eagerMaterializeBytes` = 2 GB): a NUMERIC_LONG column whose packed BODY is 0.4 GB across leaves and whose decoded 8 B/row lane is 1.6 GB projects 2.0 GB, so the fold scan leaves `retained = 0.4 GB` and the check computes 0.4 + 2.0 = 2.4 GB &gt; 2.0 GB and throws `FillBudgetExceededException` — although the store&#39;s true residency after the fill would be exactly 2.0 GB, inside the budget, because the 0.4 GB body arrays are the very same objects. The group arms answer that decline by re-entering `wholeLeafOnly = true` (SirixVectorizedExecutor.java:13121, :13399), so a servable sliced query is permanently steered to the slower whole-leaf byte-kernel scan for 0.4 GB of phantom residency. The same double count sits in the planner predicates `columnFillWithinBudget` (:769) and `columnIdentityFillable` (:1171), and in `columnDistinctIdentity`&#39;s check (:1221), so the route is rejected before the door is even reached. Never a wrong answer, but it is the exact defect class the round set out to remove (&#34;the ledger charges each retained byte exactly ONCE&#34;), fixed on the mutation side and missed on the predicate side. Fix at the same boundary: price the CHECK against the delta too — e.g. have `checkFillBudget`/`columnFillWithinBudget`/`columnIdentityFillable` subtract `projectedColumnBodyFillBytes(col)` when `columnBytes[col]` is already published, exactly as `chargeRetained` does, so the gate and the ledger agree about what a fill will add.
- ℹ️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionColumnStore.java:1024` - The `retainedFillBytes` javadoc still documents the behaviour commit 2a11405ab removed: &#34;The charge is per PUBLISH and may double-count BODY bytes shared between a raw-bytes fill and an identity fill of the same column. That direction is deliberate: over-counting declines earlier, which is the safe side of a residency cap.&#34; `chargeRetained` (:1042-1060), added in that same commit and called from both `column()` (:838) and `columnDistinctIdentity()` (:1234), now subtracts what `columnBytes` already charged specifically so the BODY chain is counted once — and its own javadoc says the over-charge &#34;steers servable queries off the sliced route for residency that is not there&#34;, i.e. the opposite verdict. A reader who reaches the field&#39;s doc first is told the ledger is deliberately over-conservative, which is the assumption that would justify leaving the check side double-counting. Rewrite the paragraph to state the charge-once rule and name `chargeRetained` as where it is enforced.

🔧 Fix: Price fill-budget checks incrementally, matching the ledger
2 issues (1 warning, 1 info) still open:
- ⚠️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:14735` - The fix rounds taught the two group-aggregate arms that a `FillBudgetExceededException` is a routing decision (catch at :13121 and :13399, re-enter `wholeLeafOnly`, tick no defect counter), but two other routes that perform unmasked whole-column fills still treat it as corruption. (1) SORTED SCAN: `sortedScanRecordKeys` gates its sliced arm on `predsSliceable` only (:14636, kind-only since round 4) and does NOT gate the ORDER columns at all; `ProjectionColumnScan.topKRecordKeys` then calls `store.column(sortColumns[k], fetcher)` (ProjectionColumnScan.java:1211), which throws the budget decline for an over-budget column. It escapes to `catch (final RuntimeException e)` at :14733 and increments `SORTED_SCAN_FAILED` (:14735). Concrete path on the advertised 8 G heap (`columnFillBudgetBytes` = `eagerMaterializeBytes` ≈ 2 GB): a 100M projection whose `URL` STRING_DICT column projects ~3.5 GB; `for $r in $doc[] order by $r.URL return $r` with a limit selects the sliced top-K arm (`anyStringKey &amp;&amp; sliced &amp;&amp; limit &gt;= 0`), the fill declines, and every such query ticks a defect counter AND falls to the generic navigational pipeline over 100M records instead of the whole-leaf windowed byte kernel the budget declined it toward. `collectMatchingSortTuples` on the same route (:14666) has the identical shape. (2) PREDICATE VALUE EMISSION: `ProjectionColumnScan.matchingFieldValues` takes `keep == null ? store.column(valueColumn, fetcher) : store.columnMasked(...)` (ProjectionColumnScan.java:892-893) — both doors now price against the budget — and the decline escapes to `catch (final RuntimeException e)` at :14571, incrementing `PREDICATE_VALUE_EMISSION_FAILED` (:14572). A CONTAINS or ordering literal on a fat STRING_DICT column reaches the null-mask arm because `computeKeepMask` prunes strings only for `Op.EQ`. This is the same defect the group arms were fixed for, left standing on the routes the fix rounds did not visit, and it contradicts the intent&#39;s &#34;declines through the store&#39;s established non-memoizing door (deliberately not marked as corruption)&#34;. Fix at the shared boundary the group arms already use: add a `catch (final ProjectionColumnStore.FillBudgetExceededException declined)` arm ahead of the generic `RuntimeException` handler at both sites, so an over-budget fill declines quietly (and, where a whole-leaf arm exists on the same route — `sortedScanRecordKeys` already has the `!sliced` byte-scan branch at :14668-14673 — re-enters it with `sliced = false` rather than dropping to the generic pipeline).
- ℹ️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionColumnStore.java:701` - `projectedColumnBodyFillBytes(col)` (:701-717) walks every leaf descriptor with a `RowGroupDescriptor.entryIndexOf` binary search per leaf, and it is NOT memoized — unlike its sibling `projectedColumnFillBytes`, which caches into the `projectedFillBytes` array (:952-960) for exactly this reason (&#34;a real sum is never 0&#34;). Round 10 moved this walk from the publish path onto the PLANNER path: `incrementalFillBytes` (:1076) calls it, and `columnFillWithinBudget` (:768) / `columnIdentityFillable` (:1191) call `incrementalFillBytes`, so `columnFillable` — invoked per gated column by `allColumnsSliceable`, `aggColumnsFillable`, `predsFillable` and `treeSliceable` on every group/aggregate query, twice more on a `wholeLeafOnly` re-entry — now redoes the walk whenever that column&#39;s raw BODY chain is already published. That is precisely the ClickBench-shaped case the change targets: a fused fold scan (`ProjectionColumnSegmentFoldScan` → `store.columnBytes`) fills the body, and every later query on the same cached handle pays ~97,654 × 5 descriptor probes per gated column. CLAUDE.md&#39;s hot-path rules call this out; the fix is mechanical — memoize it in an `AtomicReference&lt;long[]&gt;` exactly as `projectedColumnFillBytes` does, since the figure is a pure function of the immutable descriptors.

🔧 Fix: Treat budget declines as routing on every fill route
2 infos still open:
- ℹ️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:14659` - The new budget re-entry cannot serve the case it was added for, and the decline goes uncounted. `sortedScanRecordKeys`&#39;s FillBudgetExceededException handler (:14755) re-enters with `wholeLeafOnly = true`, which forces `sliced = false` (:14658); the very next gate is `if (anyStringKey &amp;&amp; !(sliced &amp;&amp; limit &gt;= 0)) return null;` (:14659), so a STRING_DICT order key returns null at the gate — before the whole-leaf collector, and before any counter. `SORTED_SCAN_DECLINED` only ticks when the second pass throws again (:14761), which cannot happen once the gate short-circuits. Concrete path (the one round 10 named): a 100M projection whose `URL` STRING_DICT column projects ~3.5 GB against the ~2 GB `columnFillBudgetBytes`; `for $r in $doc[] order by $r.URL return $r` with a limit takes the sliced top-K arm, `ProjectionColumnScan.topKRecordKeys` calls `store.column(URL, fetcher)` (:1211), the store refuses, the retry hits :14659 and returns null. Net behavior is unchanged from before this commit (the generic navigational pipeline answers) and the false `SORTED_SCAN_FAILED` tick is correctly gone, but the method&#39;s new javadoc (:14594-14598, &#34;the whole-leaf branch below … can [serve], so the decline re-enters there instead of dropping a servable query to the generic navigational pipeline&#34;) is false for this arm — the fall-through collectors carry long tuples only (:14680) — and the route change is invisible in both counters, which is the bare-null observability hole the earlier rounds asked to close. Fix: tick `SORTED_SCAN_DECLINED` at :14659 when `wholeLeafOnly` is set, and narrow the javadoc to say the whole-leaf re-entry serves long order keys only.
- ℹ️ `bundles/sirix-query/src/main/java/io/sirix/query/scan/SirixVectorizedExecutor.java:14829` - This commit re-created the orphaned-javadoc defect that round 1 reported and the author fixed in `Names`/`PathSummaryWriter`/`StorageEngineWriter`, and round 7 fixed again in `SirixVectorizedExecutor`/`ProjectionIndexCatalog`: a new member inserted between an existing doc block and the member it documented. (1) :14829 — `/** Sorted-scan serving attempts that FAILED with an exception (not gate declines). */` now precedes the newly inserted `SORTED_SCAN_DECLINED` field and its own doc block, stacking two doc comments on one member and leaving `SORTED_SCAN_FAILED` (:14837) undocumented. (2) :14867 — `/** Value-emission attempts that FAILED with an exception (not gate declines). */` now precedes `PREDICATE_VALUE_EMISSION_DECLINED`, leaving `PREDICATE_VALUE_EMISSION_FAILED` (:14875) undocumented. Generated javadoc attributes each stale block to the wrong counter, and the two blocks say opposite things about the same member (&#34;FAILED with an exception&#34; vs &#34;a route decision, not corruption&#34;). Move each new field/accessor pair below the member the preceding block belongs to.

</details>

<details>
<summary>🔧 **Test** - 1 issue found → auto-fixed (3) ✅</summary>

- ⚠️ `bundles/sirix-query/build.gradle:207` - The `projectionRefusalTest` Gradle task added by this change (bundles/sirix-query/build.gradle:207) is wired into `check` but verifies nothing: it includes `**/ProjectionRefusalPathTest.class`, and no `ProjectionRefusalPathTest` source or class exists anywhere in the repository. The system property it provides, `sirix.projection.maxIncrementalRecords`, is also read by no Java code — it appears only in build.gradle. `./gradlew :sirix-query:projectionRefusalTest --rerun-tasks` reports BUILD SUCCESSFUL and produces no test-results XML at all. The paired `exclude &#39;**/ProjectionRefusalPathTest.class&#39;` in the main `test` task is likewise a no-op. The intent says three pieces were dropped as structurally superseded, including &#34;refusal-site counters&#34;, so this is most likely vestigial build wiring left behind by that drop — but as it stands `check` gained a verification gate that can never fail. The author should decide whether to land the missing test or remove the task; I did not change build configuration myself since that is beyond a test fix.
- <code>`./gradlew :sirix-core:test --tests io.sirix.index.projection.ParallelBulkProjectionEquivalenceTest --tests io.sirix.access.trx.node.json.ParallelBulkPrimitiveIndexEquivalenceTest --tests io.sirix.index.projection.ProjectionWindowedPayloadServeTest --tests io.sirix.index.projection.ProjectionPostPassBoundedRetentionTest --tests io.sirix.index.projection.ProjectionBulkSlotAccumulationTest --tests io.sirix.index.projection.ProjectionStaleReasonTest --tests io.sirix.index.projection.ProjectionStoreKindConsistencyTest --tests io.sirix.index.projection.ProjectionSegmentResurrectionTest --tests io.sirix.index.projection.ProjectionGlobalDictionaryGenerationRotationTest --tests io.sirix.page.HOTCompleteDumpMergeTest --tests io.sirix.page.HOTTombstoneEvictionTest --tests io.sirix.index.hot.HOTBulkBuilderTest --tests io.sirix.index.hot.HOTIntegrateTest --tests io.sirix.cache.FrameSlotAllocatorTest --tests io.sirix.io.HashAccessesEquivalenceTest --tests io.sirix.access.trx.node.json.BulkAssemblyEquivalenceOracleTest` — 101 tests, 0 failures</code>
- <code>`./gradlew :sirix-core:test --tests io.sirix.index.hot.StraddleCanonicityProbe --tests io.sirix.index.hot.HOTInsertionOrderShapeProbe --tests io.sirix.index.hot.BetaIsDiscBitRoutingProbe --tests io.sirix.index.hot.HOTIndirectPageSplitFaithfulTest --tests io.sirix.access.trx.page.AsyncFlushHftPhaseTelemetryTest --tests io.sirix.index.projection.ProjectionStreamingGlobalDictionaryTest` — 24 tests, 0 failures</code>
- <code>`./gradlew :sirix-query:test --tests io.sirix.query.ProjectionIndexCatalogServingTest` — 45 tests, 0 failures</code>
- <code>`./gradlew :sirix-query:test --tests io.sirix.query.scan.SirixVectorizedExecutorLifecycleTest --tests io.sirix.query.ProjectionLoadTimeBuildEquivalenceTest --tests io.sirix.query.GlobalDictMaintenanceVerdictTest --tests io.sirix.query.ProjectionSegmentSlotMaintenanceTest --tests io.sirix.query.bench.clickbench.ClickBenchProjectionTest --tests io.sirix.query.function.jn.temporal.StoreValidTimeAutoIndexTest` — 46 tests, 0 failures</code>
- <code>`./gradlew :sirix-query:projectionRefusalTest --rerun-tasks` — BUILD SUCCESSFUL but produced no test-results XML (task matches no test class)</code>
- <code>`java ... io.sirix.query.bench.clickbench.ClickBenchGenerateMain /tmp/cbev/hits-200k.json 200000 42` — generated the 200,000-record / 390 MB synthetic ClickBench corpus used by every end-to-end run</code>
- <code>`java ... io.sirix.query.bench.clickbench.ClickBenchPrimitiveIndexImportCostMain /tmp/cbev/hits-200k.json /tmp/cbev/work 3` — bare vs indexed one-pass import, with URL PATH/NAME posting read-back</code>
- <code>`java ... io.sirix.query.bench.clickbench.ClickBenchParallelProjectionCostMain /tmp/cbev/hits-200k.json /tmp/cbev/work 3` — default dictionary election; post-pass arm aborts at distinct entry 16,385</code>
- <code>`java -Dsirix.projection.globalDict=never ... ClickBenchParallelProjectionCostMain /tmp/cbev/hits-200k.json /tmp/cbev/work 2` — three-arm bare / one-pass / post-pass timing comparison</code>
- <code>`java ... io.sirix.query.bench.clickbench.ClickBenchLoadMain /tmp/cbev/db-200k generate:200000` — one-pass 25-column projection built during the shred, dictProbes=0, validation OK</code>
- <code>`java ... io.sirix.query.bench.clickbench.ClickBenchRunMain /tmp/cbev/db-200k --tries 3 --dump /tmp/cbev/dump-all43` — all 43 ClickBench queries against the one-pass index, with RSS sampled every 250 ms across 129 retired executors</code>
- <code>`java ... ClickBenchRunMain /tmp/cbev/db-200k --queries 5,20 --tries 3 --dump ...` run three times at default, `-Dsirix.projection.eagerMaterializeBytes=8388608` and `=524288` with `-Dsirix.projection.fillDiag=true`</code>
- <code>`diff -r /tmp/cbev/dump-default /tmp/cbev/dump-squeezed` — byte-identical result dumps across a ~3 GB and a 512 KB fill budget</code>
- <code>`git status --porcelain` — worktree clean; all generated corpora, databases and result dumps removed from /tmp after capture</code>

🔧 Fix: Remove vacuous projectionRefusalTest task from sirix-query build
1 warning still open:
- ⚠️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionIndexChangeListener.java:1681` - The intent requires that &#34;the one silent load-degradation (dictionary budget breach) prints unconditionally to stderr&#34;. That notice is emitted by abandonForOversizedDictionary (ProjectionIndexChangeListener.java:1681), added by 75e0fc480, and no test in the repository exercises it — grep for GLOBAL_DICTIONARY_BUDGET_EXCEEDED finds only enum round-trip coverage in ProjectionStaleReasonTest, never the emitted line. I wrote a focused test (real ParallelBulkJsonImporter loads, tee-captured System.err, loud arm + healthy control arm) and could not reach the abandon path through the public API. With sirix.projection.globalDict=auto the election declines the column on isGlobalDictionaryWorthwhile before any mid-load breach can occur — I swept budgets of 1, 2, 4, 8, 16, 32, 64, 128, 256 and 512 MiB against 20k/100k/300k-record all-distinct corpora and every run published a live per-leaf STRING_DICT column (kind 2) with no abandon. With globalDict=always the breach is reached, but AdmissionPolicy.FAIL_CLOSED wraps it in an IllegalStateException that kills the load instead of abandoning, so the notice never prints. The two election knobs that would open the window (sirix.projection.globalDict.minEntries, .dedupFactor) are private static final Integer.getInteger(...) fields resolved at class initialization, so a test cannot set them — they would have to arrive as JVM flags via a dedicated forked Gradle test task, which is precisely the pattern you just directed me to remove from bundles/sirix-query/build.gradle. I removed my test rather than leave a failing or knob-dependent one; the worktree is clean. Your call on how to proceed: add a package-private fault-injection seam on ProjectionBulkLoad so the typed decline can be raised deterministically, relax the election constants to instance/parameter values a test can supply, or accept that this notice is verified only by the 100M acceptance runs and not by CI.
- <code>`./gradlew :sirix-query:check --dry-run` — task graph now terminates at `:sirix-query:test`, no `projectionRefusalTest` node</code>
- <code>`./gradlew :sirix-query:projectionRefusalTest --dry-run` — fails with &#34;not found in project &#39;:sirix-query&#39;&#34;, confirming the task is gone</code>
- <code>`git grep -nE &#39;projectionRefusal|ProjectionRefusalPath|sirix\.projection\.maxIncrementalRecords&#39;` — zero matches tree-wide</code>
- <code>`./gradlew :sirix-query:test --tests io.sirix.query.ProjectionIndexCatalogServingTest` — 45 tests, incl. `anOverBudgetSortedScanReEntersTheWholeLeafArmInsteadOfCountingAsCorruption` and `anOverBudgetColumnRoutesToTheWholeLeafKernelInsteadOfDecliningIntoTheSlicedArm` from the newest review commits</code>
- <code>`./gradlew :sirix-core:test --tests io.sirix.index.projection.ParallelBulkProjectionEquivalenceTest` — one-pass vs post-pass byte differential plus its `theDifferentialCanFail` negative control</code>
- <code>`./gradlew :sirix-core:test --tests io.sirix.index.projection.ProjectionWindowedPayloadServeTest` — 13 tests over windowed 128-leaf loads, cumulative fill pricing, and the non-memoizing decline door</code>
- <code>`./gradlew :sirix-core:test --tests io.sirix.index.projection.ProjectionStaleReasonTest` — machine-readable stale-tombstone reasons and their remedies</code>
- <code>`./gradlew :sirix-core:test --tests io.sirix.access.trx.node.json.ParallelBulkPrimitiveIndexEquivalenceTest` — PATH/CAS/NAME probed against a sequential-import oracle, incl. the late-binding path-class probe pinning the stale resolved-path-cache fix</code>
- <code>`./gradlew :sirix-core:test --tests io.sirix.access.trx.node.json.BulkAssemblyEquivalenceOracleTest` — 19 full-field dump differentials incl. mid-load async-flush epochs</code>
- <code>Manual end-to-end: compiled and ran `BulkIndexCampaignDemo` against the project&#39;s test runtime classpath — bulk import + `jn:create-projection-index` in one commit, cold re-open, four JSONiq queries through `SirixVectorizedExecutor`, each cross-checked against `SirixCompileChain.createWithJsonStoreWithoutAutoWiring` and reported with the catalog&#39;s served counter</code>
- <code>Attempted (removed): a focused `System.err` capture test for the dictionary-budget abandon notice; swept budgets 1 MiB→512 MiB × 20k/100k/300k records in both `auto` and `always` dictionary modes — the abandon path was unreachable through the public load API</code>

🔧 Fix: Test the unconditional dictionary-budget abandon notice
1 warning still open:
- ⚠️ `bundles/sirix-core/src/main/java/io/sirix/index/projection/ProjectionIndexChangeListener.java:1681` - The abandon notice required by the intent now prints and is properly tested, but the numbers it prints are self-contradictory. Observed verbatim from a real load: &#34;[proj] PROJECTION ABANDONED during the load: index 0, column 0 retained 1114240 B over 10227 distinct values, past its 39850930 B budget.&#34; — it reports 1.1 MB retained and calls that &#34;past&#34; a 39.8 MB budget, a number 36x larger. Cause: the budget predicate at GlobalValueDictionaryWriter.java:412-413 tests max(saturatedAdd(retainedBytes(), allocationBytes), flushPeak) &gt; budgetBytes, but refuseAdmission (GlobalValueDictionaryWriter.java:748) constructs GlobalDictionaryBudgetExceededException with retainedBytes() alone. The quantity that actually breached (the projected flush peak, or retained+allocation) is never carried and never printed. This is structural, not a quirk of my corpus: all three refusal sites (lines 413, 728, 741) compare retained PLUS a reservation against the budget, and a dictionary whose retention alone exceeded the budget would already have been refused on an earlier admission — so retained &lt; budget holds on essentially every refusal, meaning this notice can never print a coherent pair. Operator impact: the notice is the sole signal for this degradation and its own remedy text (via the exception message) says &#34;Raise the configured byte budget&#34;, but the line gives no usable number to raise it by, and reads as though the guard misfired. The new test does not catch this: NOTICE group 3 captures retained but the loud arm asserts only groups 1, 2, 4 and 5, never a retained-vs-budget relationship, so it pins the contradictory format. Fixing this means adding the breaching quantity to GlobalDictionaryBudgetExceededException and to the notice — a production change to an operator-facing output contract, beyond a test-phase fix — so it is your call whether to correct the notice (and tighten the test to assert the printed quantity actually exceeds the printed budget) or accept the current wording.
- <code>`./gradlew :sirix-core:test --tests &#39;io.sirix.index.projection.ProjectionDictionaryBudgetAbandonNoticeTest&#39; --rerun` — 2/2 passed (loud arm + healthy control arm)</code>
- <code>`./gradlew :sirix-core:test --tests &#39;io.sirix.index.projection.ProjectionStaleReasonTest&#39; --rerun` — 5/5 passed (machine-readable stale-reason half of the same intent clause)</code>
- <code>Reachability trace of `abandonForOversizedDictionary`: callers at `ProjectionIndexChangeListener.java:1650` (`beforeCommit`) and `:1708` (`beforePageFlush`); decline raised at `GlobalValueDictionaryWriter.java:748`; budget predicate at `GlobalValueDictionaryWriter.java:412-413`</code>
- <code>Mutation 1 (non-vacuity): suppressed the notice via `if (false) System.err.println(...)` at `ProjectionIndexChangeListener.java:1681` → loud arm FAILED (`expected: &lt;1&gt; but was: &lt;0&gt;`), control arm still passed; reverted</code>
- <code>Mutation 2 (unsuppressible-channel contract): rerouted the identical text from `System.err.println` to `LOGGER.warn` → loud arm FAILED with empty captured stderr under `Level.OFF`, control arm still passed; reverted</code>
- `Captured the verbatim operator-facing stderr line emitted by a real 30,000-record shredder load with the root logger set to OFF`
- <code>Verified `git status --porcelain` is empty after both mutations were reverted — no production or test source left modified</code>

🔧 Fix: Print the breaching quantity in the dictionary abandon notice
✅ Re-checked - no issues remain.
- <code>`./gradlew :sirix-core:test --rerun --tests &#39;io.sirix.index.projection.ProjectionDictionaryBudgetAbandonNoticeTest&#39; --tests &#39;io.sirix.index.projection.GlobalValueDictionaryWriterBudgetTest&#39; --tests &#39;io.sirix.index.projection.ProjectionStaleReasonTest&#39;` — 20 tests, 0 failures; captured the real operator stderr line from a live 30k-record load</code>
- <code>Regression A/B: `git checkout HEAD~1 -- bundles/sirix-core/src/main/java/io/sirix/index/projection/{GlobalDictionaryBudgetExceededException,GlobalValueDictionaryProbeFront,GlobalValueDictionaryWriter,ProjectionIndexChangeListener}.java` then re-ran `ProjectionDictionaryBudgetAbandonNoticeTest` — loud arm FAILS on the pre-fix notice (AssertionFailedError at ProjectionDictionaryBudgetAbandonNoticeTest.java:163), passes after restoring the fix. Production files restored.</code>
- <code>`./gradlew :sirix-core:test --rerun --tests &#39;io.sirix.index.projection.ProjectionGlobalDictionaryGenerationRotationTest&#39;` — the FAIL_CLOSED structural-ceiling wrapper path still matches its message assertion (`mentionsTheCeiling`), unaffected by the two-factory refactor</code>
- <code>Test improvement: extended `boundedWriterRefusesAtTheBudget` to assert the byte-decline invariant on the exception (term named, `breachingBytes &gt; budgetBytes`, `retained &lt;= breaching`, message states its own arithmetic, remedy quotes a raisable budget)</code>
- <code>Test improvement: added `GlobalValueDictionaryWriterBudgetTest &gt; &#34;A structural ceiling declines without claiming any quantity exceeded the byte budget&#34;` — drives the real writer into the entry-count and value-length ceilings and asserts `breachingTerm() == null`, no &#34;past the&#34; breach claim, and the structural remedy text</code>
- <code>Manual verification: drove the real `GlobalValueDictionaryWriter` to each decline family through its own API (no fault injection) and printed the operator-facing message for both, confirming the byte-budget message is self-consistent and the structural message invents no comparison</code>
- <code>Cleanup verified: `git status --short --untracked-files=all` shows only the intentional test-file modification; scratch verification program and extracted classpath removed from /tmp</code>

</details>

<details>
<summary>🔧 **Document** - 2 issues found → auto-fixed (2) ✅</summary>

- ℹ️ `docs/HOT_BULK_BUILD.md:3` - docs/HOT_BULK_BUILD.md (546 lines) is a design record pinned to base commit 226ceaac8, and this change landed most of it. I removed the contradictions in place (status header, the §0 premise bullet, Seam 1&#39;s &#39;currently refuses&#39;, Seam 2b&#39;s &#39;uncommitted&#39; heading, the Appendix answer) and pointed shipped behavior at BULK_IMPORT.md, rather than rewriting the document. A follow-up could trim it to the reasoning that is still load-bearing and let BULK_IMPORT.md own everything about what the loaders do today; that consolidation is larger than this change made stale, so I left it out.
- ℹ️ `docs/PROJECTION_INDEXES.md:147` - The code defines ~50 `sirix.projection.*` system properties; only six appear in any document, scattered across five files, and no document owns the projection tuning surface. I documented just the one this change made user-visible (`sirix.projection.eagerMaterializeBytes`, in PROJECTION_INDEXES.md, matching where projection knobs already live — operations.md deliberately owns only the `sirix.cache.*` budgets). A durable fix is a generated property table checked for drift against the source, which is out of scope here.

🔧 Fix: Consolidate HOT bulk design record into reasoning-only owner
2 infos still open:
- ℹ️ `docs/HOT_BULK_BUILD.md:1` - Judgment call worth confirming: I deleted the §2.1 leaf-packing narrative (the measured 280→40 leaf improvement) because the operative invariant is owned by HOTBulkBuilder.isExpandable/fitsLeafPage javadoc and the placement policy forbids incident narratives in prose. But the deleted text was the only place asserting the outcome — HOTInsertionOrderShapeProbe REPORTS leaf counts and asserts only read-equivalence (probe line 56: &#34;Shape equality is REPORTED, not asserted&#34;), and HOTBulkBuilderTest asserts leaf counts only for determinism (line 248, ra vs rb) and the single-leaf case (line 210). So the highest-fitting-subtree packing policy currently has no regression guard: a future frontier change could silently reintroduce the ~7x fragmentation and every existing gate would stay green. Adding an assertion is a test change, out of scope for this documentation/lint phase.
- ℹ️ `docs/PROJECTION_INDEX_DEEP_DIVE.md:631` - Judgment call on where the descriptive material went. The instruction said to move descriptive content to BULK_IMPORT.md, but one block did not belong there: the bulk slot sink is engaged by the POST-PASS projection build (jn:create-projection-index via ProjectionIndexBuilder), not by either JSON bulk loader — the load-time rider ProjectionBulkLoad deliberately still writes per entry. Filing it under BULK_IMPORT.md would have made that document the owner of a fact about a code path it does not cover, against the one-owner rule. I routed it to PROJECTION_INDEX_DEEP_DIVE.md §6.1 (the projection write path&#39;s owner) instead, and pointed HOT_BULK_BUILD.md there. Loader-side descriptive material (residency limit, the postings-vs-bytes gate correction) did go to BULK_IMPORT.md as directed.

🔧 Fix: Guard HOT leaf-packing invariant; mutation-proven by reverting fit-bound
✅ Re-checked - no issues remain.

</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,23 @@ All notable changes to SirixDB are documented in this file.
   so nested columns sharing a trailing name with another path are accepted. See
   `docs/PROJECTION_INDEXES.md` and `docs/PROJECTION_INDEX_INCREMENTAL_MAINTENANCE.md`.
 - **Load-time projection builds** — `BasicJsonDBStore#create(collection, resource, reader|parser,
-  ProjectionSpec)` catalogues a projection on the still-empty resource and lets the shred's own
-  change notifications fill it, so a load plus its projection is ONE pass instead of a shred
-  followed by a full `jn:create-projection-index` walk. Until the load's final commit the
-  projection's metadata slot holds the stale tombstone, so an interrupted load leaves queries on
-  the generic pipeline rather than on a half-filled index. See `docs/PROJECTION_INDEXES.md`.
+  ProjectionSpec)` catalogues a projection on the still-empty resource and lets the load fill it,
+  so a load plus its projection is ONE pass instead of a shred followed by a full
+  `jn:create-projection-index` walk. Until the load's final commit the projection's metadata slot
+  holds the stale tombstone, so an interrupted load leaves queries on the generic pipeline rather
+  than on a half-filled index. See `docs/PROJECTION_INDEXES.md`.
+- **Bulk JSON loaders for fresh resources** — `BulkJsonTreeAssembler` (sequential) and
+  `ParallelBulkJsonImporter` (feeder scan + worker page builders + ordered adoption, for corpora
+  whose top level is an array; NDJSON rides `NdjsonAsArrayInputStream`). Both build trees
+  structurally identical to cursor-based insertion, enforced by a full-field differential oracle,
+  and refuse up front what they do not reproduce faithfully (`hashType` other than `NONE`, stored
+  DeweyIDs, node history, path statistics, a non-empty target). The parallel importer **maintains
+  PATH, CAS and NAME definitions and armed projection builds in the load's single pass** — the
+  workers extract each family's tuples from the primitives they already hold and the coordinator
+  drains them into the families' ordinary bulk loaders; a catalogued projection with no armed
+  load-time build is refused rather than silently left unmaintained, and valid-time interval
+  maintenance is the one family that still refuses. API, scope, verification, tuning and measured
+  numbers: `docs/BULK_IMPORT.md`.
 - **Asynchronous durable commits** (`AfterCommitState.KEEP_OPEN_ASYNC_COMMIT`) — the middle
   ground between synchronous auto-commits and the async pre-flush: every threshold crossing
   creates a real, durable, queryable revision, but the durability barriers (index-catalogue
@@ -45,6 +57,21 @@ All notable changes to SirixDB are documented in this file.
 
 ### Changed
 
+- **Wide projection string columns serve through windowed leaf loads** instead of whole-column
+  eager materialization. A projection whose worst-case resident size exceeds
+  `-Dsirix.projection.eagerMaterializeBytes` (default: the smaller of half the projection cache
+  budget and a quarter of the heap) is served from bounded 128-leaf windows, and a column fill
+  that would exceed the budget declines — the query re-enters the windowed whole-leaf route where
+  one exists, otherwise the record path answers. A decline is a routing decision, not a corruption
+  signal, so the index stays valid. Whole-column materialization is unchanged below the budget.
+  See `docs/PROJECTION_INDEXES.md`.
+- **A projection abandoned mid-load reports unconditionally.** The one silent load-degradation —
+  a resource-wide value dictionary breaching its byte budget, which abandons the projection while
+  the load still succeeds — now prints `[proj] PROJECTION ABANDONED` on stderr with the quantity
+  that breached the budget (the shipped log configuration discards the warning), and the stale
+  tombstone records a machine-readable `StaleReason` plus its remedy. The reason rides previously
+  reserved flag bits, so the `PIXM` wire format is unchanged for existing tombstones
+  (`docs/DISK_FORMAT.md`).
 - **JSON revision-diff sidecars carry an integrity envelope** (`sirix-diff-format`,
   `operation-count`, `operations-sha256`). Readers (`jn:diff`, the REST diff handler,
   multi-revision resource copy) validate identity, count and digest before use and fall back to
--- a/bundles/sirix-core/build.gradle
+++ b/bundles/sirix-core/build.gradle
@@ -129,7 +129,16 @@ test {
             excludeTags 'heavy'
         }
     }
-    
+
+    // HOT fragment-merge / eviction / split counters are OFF in production: they sit on the default
+    // read path and inside a per-entry commit loop, so they are gated behind a static-final flag
+    // that folds the branch away entirely. The suites asserting on them need the flag ON — and they
+    // ASSERT it is on rather than trusting it, because with the gate off every counter reads zero
+    // and "zero walked past a complete dump" from a disabled instrument is indistinguishable from
+    // the same reading from a healthy one. This line is the "provide" half of assert-and-provide.
+    // An explicit -Dsirix.hot.mergeDiag=false still wins, so the gated-off path stays testable.
+    systemProperty 'sirix.hot.mergeDiag', System.getProperty('sirix.hot.mergeDiag', 'true')
+
     // Add async-profiler if ASYNC_PROFILER env var is set
     // Usage: ASYNC_PROFILER=/path/to/async-profiler PROFILE_EVENT=cpu|alloc PROFILE_OUTPUT=/tmp/profile.txt ./gradlew test
     def profilerPath = System.getenv("ASYNC_PROFILER")
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractNodeTrxImpl.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractNodeTrxImpl.java
@@ -733,6 +733,35 @@ public abstract class AbstractNodeTrxImpl<R extends NodeReadOnlyTrx & NodeCursor
   }
 
   /**
+   * Bulk-assembly epoch driver: account a whole completed record's mutations in one call and rotate
+   * the intermediate epoch under EXACTLY the same predicate the per-insert path uses — same count
+   * threshold, same TIL page-work boundary, same flush dispatch. The assembler computes its pointers
+   * from a stack instead of per-insert bookkeeping, so it accounts at record boundaries; the only
+   * semantic difference to the cursor path is that rotation happens AFTER accounting the
+   * just-completed record, so an epoch may exceed the threshold by at most one record's node count
+   * (records are orders of magnitude smaller than an epoch).
+   *
+   * @param mutations completed node creations since the previous call; must be positive
+   */
+  protected final void bulkAccountMutations(final int mutations) {
+    if (mutations <= 0) {
+      throw new IllegalArgumentException("mutations must be positive: " + mutations);
+    }
+    assertNotRollbackOnly();
+    mutationSequence += mutations;
+    modificationCount += mutations;
+    if (shouldRotateIntermediateEpoch()) {
+      if (afterCommitState == AfterCommitState.KEEP_OPEN_ASYNC_COMMIT) {
+        asyncCommitInternal("autoCommit");
+      } else if (afterCommitState == AfterCommitState.KEEP_OPEN_ASYNC_FLUSH) {
+        flushIntermediateAsyncEpoch();
+      } else {
+        commitInternal("autoCommit", null, true);
+      }
+    }
+  }
+
+  /**
    * Shared insertion preflight. A subtree shredder already established that the transaction is open
    * and running before enabling bulk mode, so its per-node inserts retain count-based rotation while
    * avoiding two redundant state checks. Standalone JSON/XML insertions keep the full public-mutator
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
@@ -968,50 +968,53 @@ public abstract class AbstractResourceSession<R extends NodeReadOnlyTrx & NodeCu
   @Override
   public synchronized void close() {
     if (!isClosed) {
-      // NOTE: ClockSweepers are GLOBAL now - don't stop them per-session
-      // They continue running, managed by BufferManager lifecycle
+      try {
+        // NOTE: ClockSweepers are GLOBAL now - don't stop them per-session
+        // They continue running, managed by BufferManager lifecycle
 
-      // Close all shared per-thread read-only transactions.
-      for (final NodeReadOnlyTrx rtx : sharedTrxMap.values()) {
-        rtx.close();
-      }
-      sharedTrxMap.clear();
-
-      // Close all open node transactions.
-      for (NodeReadOnlyTrx rtx : nodeTrxMap.values()) {
-        if (rtx instanceof XmlNodeTrx xmlNodeTrx) {
-          xmlNodeTrx.rollback();
-        } else if (rtx instanceof JsonNodeTrx jsonNodeTrx) {
-          jsonNodeTrx.rollback();
+        // Close all shared per-thread read-only transactions.
+        for (final NodeReadOnlyTrx rtx : sharedTrxMap.values()) {
+          rtx.close();
         }
-        rtx.close();
-      }
-      // Close all open storage engine writers.
-      for (StorageEngineReader rtx : storageEngineWriterMap.values()) {
-        rtx.close();
-      }
-      // Close all open storage engine readers.
-      for (StorageEngineReader rtx : storageEngineReaderMap.values()) {
-        rtx.close();
-      }
+        sharedTrxMap.clear();
 
-      // NOTE: Don't clear BufferManager caches here - other sessions might be using same resource!
-      // Pages will be evicted by normal cache LRU policy or cleaned up at database close
-      // PostgreSQL-style: buffers released when ALL sessions release them, not when one closes
+        // Close all open node transactions.
+        for (NodeReadOnlyTrx rtx : nodeTrxMap.values()) {
+          if (rtx instanceof XmlNodeTrx xmlNodeTrx) {
+            xmlNodeTrx.rollback();
+          } else if (rtx instanceof JsonNodeTrx jsonNodeTrx) {
+            jsonNodeTrx.rollback();
+          }
+          rtx.close();
+        }
+        // Close all open storage engine writers.
+        for (StorageEngineReader rtx : storageEngineWriterMap.values()) {
+          rtx.close();
+        }
+        // Close all open storage engine readers.
+        for (StorageEngineReader rtx : storageEngineReaderMap.values()) {
+          rtx.close();
+        }
 
-      // Immediately release all resources.
-      nodeTrxMap.clear();
-      storageEngineReaderMap.clear();
-      storageEngineWriterMap.clear();
-      ProjectionIndexCatalog.invalidateSessionWindowedPayloads(this);
-      resourceStore.closeResourceSession(resourceConfig.getResource());
+        // NOTE: Don't clear BufferManager caches here - other sessions might be using same resource!
+        // Pages will be evicted by normal cache LRU policy or cleaned up at database close
+        // PostgreSQL-style: buffers released when ALL sessions release them, not when one closes
 
-      storage.close();
+        // Immediately release all resources.
+        nodeTrxMap.clear();
+        storageEngineReaderMap.clear();
+        storageEngineWriterMap.clear();
+        resourceStore.closeResourceSession(resourceConfig.getResource());
 
-      if (pool.get() != null) {
-        pool.get().close();
+        storage.close();
+
+        if (pool.get() != null) {
+          pool.get().close();
+        }
+        isClosed = true;
+      } finally {
+        ProjectionIndexCatalog.invalidateSessionWindowedPayloads(this);
       }
-      isClosed = true;
     }
   }
 
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
@@ -33,6 +33,7 @@ import io.sirix.exception.SirixThreadedException;
 import io.sirix.exception.SirixUsageException;
 import io.sirix.index.IndexType;
 import io.sirix.index.path.summary.PathSummaryReader;
+import io.sirix.index.projection.ProjectionIndexCatalog;
 import io.sirix.io.IOStorage;
 import io.sirix.io.Reader;
 import io.sirix.io.RevisionIndex;
@@ -1002,6 +1003,7 @@ public abstract class AbstractResourceSession<R extends NodeReadOnlyTrx & NodeCu
       nodeTrxMap.clear();
       storageEngineReaderMap.clear();
       storageEngineWriterMap.clear();
+      ProjectionIndexCatalog.invalidateSessionWindowedPayloads(this);
       resourceStore.closeResourceSession(resourceConfig.getResource());
 
       storage.close();
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AbstractResourceSession.java
@@ -33,7 +33,6 @@ import io.sirix.exception.SirixThreadedException;
 import io.sirix.exception.SirixUsageException;
 import io.sirix.index.IndexType;
 import io.sirix.index.path.summary.PathSummaryReader;
-import io.sirix.index.projection.ProjectionIndexCatalog;
 import io.sirix.io.IOStorage;
 import io.sirix.io.Reader;
 import io.sirix.io.RevisionIndex;
@@ -968,53 +967,49 @@ public abstract class AbstractResourceSession<R extends NodeReadOnlyTrx & NodeCu
   @Override
   public synchronized void close() {
     if (!isClosed) {
-      try {
-        // NOTE: ClockSweepers are GLOBAL now - don't stop them per-session
-        // They continue running, managed by BufferManager lifecycle
+      // NOTE: ClockSweepers are GLOBAL now - don't stop them per-session
+      // They continue running, managed by BufferManager lifecycle
 
-        // Close all shared per-thread read-only transactions.
-        for (final NodeReadOnlyTrx rtx : sharedTrxMap.values()) {
-          rtx.close();
-        }
-        sharedTrxMap.clear();
-
-        // Close all open node transactions.
-        for (NodeReadOnlyTrx rtx : nodeTrxMap.values()) {
-          if (rtx instanceof XmlNodeTrx xmlNodeTrx) {
-            xmlNodeTrx.rollback();
-          } else if (rtx instanceof JsonNodeTrx jsonNodeTrx) {
-            jsonNodeTrx.rollback();
-          }
-          rtx.close();
-        }
-        // Close all open storage engine writers.
-        for (StorageEngineReader rtx : storageEngineWriterMap.values()) {
-          rtx.close();
-        }
-        // Close all open storage engine readers.
-        for (StorageEngineReader rtx : storageEngineReaderMap.values()) {
-          rtx.close();
-        }
-
-        // NOTE: Don't clear BufferManager caches here - other sessions might be using same resource!
-        // Pages will be evicted by normal cache LRU policy or cleaned up at database close
-        // PostgreSQL-style: buffers released when ALL sessions release them, not when one closes
-
-        // Immediately release all resources.
-        nodeTrxMap.clear();
-        storageEngineReaderMap.clear();
-        storageEngineWriterMap.clear();
-        resourceStore.closeResourceSession(resourceConfig.getResource());
-
-        storage.close();
-
-        if (pool.get() != null) {
-          pool.get().close();
-        }
-        isClosed = true;
-      } finally {
-        ProjectionIndexCatalog.invalidateSessionWindowedPayloads(this);
+      // Close all shared per-thread read-only transactions.
+      for (final NodeReadOnlyTrx rtx : sharedTrxMap.values()) {
+        rtx.close();
       }
+      sharedTrxMap.clear();
+
+      // Close all open node transactions.
+      for (NodeReadOnlyTrx rtx : nodeTrxMap.values()) {
+        if (rtx instanceof XmlNodeTrx xmlNodeTrx) {
+          xmlNodeTrx.rollback();
+        } else if (rtx instanceof JsonNodeTrx jsonNodeTrx) {
+          jsonNodeTrx.rollback();
+        }
+        rtx.close();
+      }
+      // Close all open storage engine writers.
+      for (StorageEngineReader rtx : storageEngineWriterMap.values()) {
+        rtx.close();
+      }
+      // Close all open storage engine readers.
+      for (StorageEngineReader rtx : storageEngineReaderMap.values()) {
+        rtx.close();
+      }
+
+      // NOTE: Don't clear BufferManager caches here - other sessions might be using same resource!
+      // Pages will be evicted by normal cache LRU policy or cleaned up at database close
+      // PostgreSQL-style: buffers released when ALL sessions release them, not when one closes
+
+      // Immediately release all resources.
+      nodeTrxMap.clear();
+      storageEngineReaderMap.clear();
+      storageEngineWriterMap.clear();
+      resourceStore.closeResourceSession(resourceConfig.getResource());
+
+      storage.close();
+
+      if (pool.get() != null) {
+        pool.get().close();
+      }
+      isClosed = true;
     }
   }
 
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AfterCommitState.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/AfterCommitState.java
@@ -49,6 +49,14 @@ public enum AfterCommitState {
    * page-work bound. This second bound limits foreground index/projection maintenance even when a
    * workload modifies relatively few storage pages.
    * </p>
+   *
+   * <p>
+   * Overridable for measurement via {@code -Dsirix.asyncFlush.maxNodeCount}. The default keeps the
+   * historical bound; the ClickBench hits shape it implies is an epoch every ~78 rows, which makes a
+   * PARTITIONED parallel ingest submit a flush every few milliseconds per writer and starve on the
+   * global append executor's admissions (measured: 16 writers spent 76% of a 1M-row load in
+   * submitWait). A larger epoch trades bounded in-flight frozen-TIL memory for admission pressure.
+   * </p>
    */
-  public static final int MAX_ASYNC_FLUSH_NODE_COUNT = 1 << 14;
+  public static final int MAX_ASYNC_FLUSH_NODE_COUNT = Integer.getInteger("sirix.asyncFlush.maxNodeCount", 1 << 14);
 }
--- a/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/json/BulkJsonScanner.java
+++ b/bundles/sirix-core/src/main/java/io/sirix/access/trx/node/json/BulkJsonScanner.java
@@ -1,0 +1,383 @@
+/*
+ * Copyright (c) 2026, SirixDB. All rights reserved.
+ */
+package io.sirix.access.trx.node.json;
+
+import io.sirix.service.json.JsonNumber;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Minimal strict-JSON scanner for the bulk assembler: one forward pass over a {@link Reader},
+ * emitting structural events without the general-purpose tokenizer overhead the profile charged at
+ * ~25% of fill time — no per-token enum peek state machine, no {@code String} materialization for
+ * values (UTF-8 bytes are encoded straight from the char buffer into a reusable array), and field
+ * names resolve through an intern table so each distinct name allocates exactly once.
+ *
+ * <p>
+ * EQUIVALENCE CONTRACT (oracle-enforced: the cursor arm still parses with Gson, so every
+ * differential fixture doubles as a tokenizer-equivalence test):
+ * <ul>
+ * <li>Escape decoding matches JSON exactly — {@code \" \\ \/ \b \f \n \r \t \/uXXXX} including
+ * surrogate pairs assembled from consecutive escapes.</li>
+ * <li>UTF-8 value bytes are byte-identical to {@code String.getBytes(UTF_8)} over the decoded
+ * chars, INCLUDING the {@code '?'} replacement byte for unpaired surrogates.</li>
+ * <li>Numbers are delegated verbatim to {@link JsonNumber#stringToNumber(String)} — identical
+ * {@code Number} types and values by construction.</li>
+ * </ul>
+ *
+ * <p>
+ * Single-threaded; the value/name buffers are reused between events, so callers must consume an
+ * event's data before requesting the next.
+ */
+final class BulkJsonScanner {
+
+  static final int EVENT_BEGIN_OBJECT = 0;
+  static final int EVENT_END_OBJECT = 1;
+  static final int EVENT_BEGIN_ARRAY = 2;
+  static final int EVENT_END_ARRAY = 3;
+  static final int EVENT_NAME = 4;
+  static final int EVENT_STRING = 5;
+  static final int EVENT_NUMBER = 6;
+  static final int EVENT_TRUE = 7;
+  static final int EVENT_FALSE = 8;
+  static final int EVENT_NULL = 9;
+  static final int EVENT_END_DOCUMENT = 10;
+
+  private static final int DEFAULT_BUFFER_CHARS = 1 << 16;
+
+  private final Reader in;
+  private final char[] buffer;
+  private int position;
+  private int limit;
+
+  /** Decoded chars of the current string/number token (escape-processed). */
+  private char[] token = new char[256];
+  private int tokenLength;
+
+  /**
+   * Exposed state of the CURRENT event vs scratch state of a scanned-ahead event. One-event
+   * lookahead must never clobber what the caller has not yet consumed: the aliasing bug this
+   * structure exists for was a peeked STRING re-encoding over the byte buffer of the string the
+   * factory was about to write — consecutive array strings shifted by one, pinned by the oracle's
+   * lone-surrogate fixture. {@link #next()} promotes scratch to current by SWAPPING the byte
+   * arrays, so the zero-copy property survives.
+   */
+  private byte[] utf8 = new byte[512];
+  private int utf8Length;
+  private byte[] scratchUtf8 = new byte[512];
+  private int scratchUtf8Length;
+
+  /** Canonical name strings, one allocation per distinct field name. */
+  private final Map<String, String> nameInterns = new HashMap<>();
+
+  private String currentName;
+  private Number currentNumber;
+  private String scratchName;
+  private Number scratchNumber;
+
+  private int peekedEvent = -1;
+
+  BulkJsonScanner(final Reader in) {
+    this(in, DEFAULT_BUFFER_CHARS);
+  }
+
+  /** Buffer size is a test seam: tiny buffers force refills inside tokens. */
+  BulkJsonScanner(final Reader in, final int bufferChars) {
+    this.in = in;
+    this.buffer = new char[Math.max(16, bufferChars)];
+  }
+
+  int peek() throws IOException {
+    if (peekedEvent < 0) {
+      peekedEvent = advance();
+    }
+    return peekedEvent;
+  }
+
+  int next() throws IOException {
+    final int event;
+    if (peekedEvent >= 0) {
+      event = peekedEvent;
+      peekedEvent = -1;
+    } else {
+      event = advance();
+    }
+    // Promote the scanned event's scratch state to the exposed slots. Swapping the arrays keeps
+    // the encoding zero-copy while guaranteeing a later peek can never touch what we return now.
+    final byte[] previouslyExposed = utf8;
+    utf8 = scratchUtf8;
+    utf8Length = scratchUtf8Length;
+    scratchUtf8 = previouslyExposed;
+    currentName = scratchName;
+    currentNumber = scratchNumber;
+    return event;
+  }
+
+  /** Valid after {@link #next()} returned {@link #EVENT_NAME}; canonical (interned) instance. */
+  String name() {
+    return currentName;
+  }
+
+  /** Valid after {@link #next()} returned {@link #EVENT_NUMBER}. */
+  Number number() {
+    return currentNumber;
+  }
+
+  /** Valid after {@link #next()} returned {@link #EVENT_STRING}; reused buffer. */
+  byte[] stringUtf8() {
+    return utf8;
+  }
+
+  int stringUtf8Length() {
+    return utf8Length;
+  }
+
+  // ==== scanning ==============================================================================
+
+  private int advance() throws IOException {
+    while (true) {
+      final int c = nextNonWhitespace();
+      switch (c) {
+        case -1 -> {
+          return EVENT_END_DOCUMENT;
+        }
+        case '{' -> {
+          return EVENT_BEGIN_OBJECT;
+        }
+        case '}' -> {
+          return EVENT_END_OBJECT;
+        }
+        case '[' -> {
+          return EVENT_BEGIN_ARRAY;
+        }
+        case ']' -> {
+          return EVENT_END_ARRAY;
+        }
+        case ',', ':' -> {
+          // Structural separators carry no event; strictness (matching them to context) is the
+          // parser-level concern the assembler's own stack enforces structurally.
+        }
+        case '"' -> {
+          scanQuoted();
+          if (peekNonWhitespaceIsColon()) {
+            scratchName = intern(token, tokenLength);
+            return EVENT_NAME;
+          }
+          encodeUtf8();
+          return EVENT_STRING;
+        }
+        case 't' -> {
+          expect("rue");
+          return EVENT_TRUE;
+        }
+        case 'f' -> {
+          expect("alse");
+          return EVENT_FALSE;
+        }
+        case 'n' -> {
+          expect("ull");
+          return EVENT_NULL;
+        }
+        default -> {
+          if (c == '-' || (c >= '0' && c <= '9')) {
+            scanNumber((char) c);
+            scratchNumber = JsonNumber.stringToNumber(new String(token, 0, tokenLength));
+            return EVENT_NUMBER;
+          }
+          throw new IOException("unexpected character '" + (char) c + "' in JSON input");
+        }
+      }
+    }
+  }
+
+  private int nextNonWhitespace() throws IOException {
+    while (true) {
+      if (position == limit && !fill()) {
+        return -1;
+      }
+      final char c = buffer[position++];
+      if (c != ' ' && c != '\t' && c != '\n' && c != '\r') {
+        return c;
+      }
+    }
+  }
+
+  /** After a closing quote: does a colon follow (name) or not (string value)? Consumes nothing. */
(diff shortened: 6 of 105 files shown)


JohannesLichtenberger:
Closing at the author's request; not for review. The work is preserved locally.

